PR #18, opened 2021-08-26
major t:wayland improvements

ghost:
merging the pr once everything is good

all existing issues and pull requests will be transferred

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,3 @@ License
 =======
 Released under the [GPLv3 license](https://www.gnu.org/licenses/gpl-3.0.html).
 
-Wiki
-====
-You can now get complete info on how to run GUI apps in Termux-Wayland [here](https://github.com/suhan-paradkar/termux-wayland/wiki/Launching-GUI-applications)


twaik:
> Yes... Please do it👍👍

That is a bad idea. T::X11 is called so because Xorg is the only possible client.

WMCB-Tech:
> > Yes... Please do it👍👍
> 
> That is a bad idea. T::X11 is called so because Xorg is the only possible client.

@twaik maybe you're right, perhaps none of the Termux applications was compiled with wayland support

twaik:
I know I am right. T::X11's architecture was built to be the minimal possible server that can handle Xwayland connection. As simple as it can be. And this server used a few tricks to handle Android behaviour.

WMCB-Tech:
maybe... what do you think @suhan-paradkar?

suhan-paradkar:
Yes... But we may need to ask @Grimler91 to change the repo name also... Among others...

Grimler91:
Back to termux-x11 it is then!

GrpeApple:
All is not changed in https://github.com/termux/termux-x11/search?q=wayland ?

twaik:
@GrpeApple Most of search results are parts of dependencies, no need to change them.

PR #19, opened 2021-08-27
Customizable Extra-Keys

suhan-paradkar:


suhan-paradkar:
> _Device name?

Oppo A12  not rooted

> Have you ever tried in addition to Termux and T:X11/Wayland also to use VNC?

Yes.. ofcourse.. I started to continue development here as I felt that VNC is very clunky...... 

> Without Configure Event only the floating KB is useful for T:X11

Well.. that is what is being implemented here... Although the goal is to use the same set of Extra-keys as defined in `.termux/termux.properties` by the user

RalfWerner:
I got lost for your device and was surprised that you can use Pop View (PiP see second 10 to 18 in the [Video](https://user-images.githubusercontent.com/45426380/125047883-b241a300-e09f-11eb-9952-09bd1f22f905.mp4)). In this mode, the Picture/Pop-View with EK-X (T:X11) can be moved up from the lower overlap area (at least Nav bar - see shot below) and remains visible with OS-KB. The _Move Events_ of the first 3 seconds and KB events (if one of the two windows is selected) also works in this case, as well as any change of the T:X11 size - not in your PiP.
If I turn _**on your** PiP (Wayland) **Choice**_, nothing happens on this device (Android 11) and on my Huawei (Android 8) only **your** PIP (no events) is possible. Both [see here](https://github.com/WMCB-Tech/termux-wayland/discussions/18#discussioncomment-1094511). Except for the effect in the **addendum** I found **no advantages** - Split is much better. But there are also a few bad side effects (possibly only on my device).

Do you possibly have an additional app, which allow you repeat the _PiP behavior_ from the [Video](https://user-images.githubusercontent.com/45426380/125047883-b241a300-e09f-11eb-9952-09bd1f22f905.mp4) (should be possible with all Android-11 devices)?

On my Gigaset (Android 9) neither Split nor PiP works, but floating OS-KB and AVD works like my Huawei.
At the same time VNC and the two T:-apps on _one_ screen, I could reach _only_ on the SM-P610 (shot here with **three EK**)
grafik

agnostic-apollo:
You can use [Taskerbar](https://play.google.com/store/apps/details?id=com.farmerbb.taskbar) app on android 7+ to launch [freeform mode](https://developer.android.com/guide/topics/ui/multi-window#overview) to make apps into resizable windows, like on desktop, its similar to PiP.

agnostic-apollo:
Use `act.findViewById()` and `act.getContext()`. Those functions are inherited from the `Activity` class and will work directly in `MainActivity` but you need to use an `Activity` object if calling them in a different class.

And no need for temporary variable [here](https://github.com/termux/termux-x11/pull/19/files#diff-bb25c1317766bc7588905f25ebdd15ce05d7dc2ddbc51a014ad582d17c2448faR110) and can just create a function that converts `char[]` to events and calls `mEventListener.onKey()`, you are using 2 loops. This is just a code improvement though.

suhan-paradkar:
Thanks👍👍

agnostic-apollo:
Sorry, `getContext()` is from `View`, Just use `act` instead of `act.getContext()`.

suhan-paradkar:
Well.. we have a new problem.... The Extra-keys are now not being displayed at all....

RalfWerner:
Thanks @agnostic-apollo for the _Taskbar_ hint and back naming in **T:X11**. I've the Taskbar-App checked and can confirm that for some apps Floating Pictures are possible. **T:X11** and **VNC** (expected) are _not_ part of it and Termux is very limited. A behavior as in the [Video](https://user-images.githubusercontent.com/45426380/125047883-b241a300-e09f-11eb-9952-09bd1f22f905.mp4) is therefore not available. The task bar is displaceable to the four corners, but almost always covers important control areas -

see _**Shot1**_. This is my Gigaset with the smallest screen, without Split/PiP from Android. With tap on **T:X11** all other apps and OS-KB disappear, **EK-X** _is covered by task bar_ and the Fluxbox task bar is _covered by **EK-X**_. I tried to work on an AVD pad with Taskbar-App, but this _completely failed._ Although **T:X11** is floating in _**Shot2**_, but only because I have ceased the short moment in which the flickering screen indicated it.
The upper bar of **Termux** and **T:X11** Picture (to move) disappears after the first touch and occures only after App restart. The Taskbar app is very instable, often crashes and sometimes can only be remedied only by device restart.
grafik

Conclusion: This PiP function is unsuitable on all devices. On small devices is _Split_ (if possible) is a good alternative but Toggle-EK in any case. [PiP](https://github.com/RalfWerner/integrated-process/issues/2#issuecomment-733700709) on my SM-P610 is a brilliant and very stable Android function.

So again @suhan-paradkar:_How did you repeat the **video behavior on your Oppo A12** (only "three fingers split" according to Google)?_

Ps [last artifact](https://github.com/termux/termux-x11/pull/19#issuecomment-908102930) ist increased from 7Mb to 10 Mb but _**not suitable for further EK-X-checks?**_

suhan-paradkar:
> How did you repeat the video behavior on your Oppo A12 (only "three fingers split" according to Google)?

Well.. pressing the home button or recents will put the app in PIP mode amd I had no trouble with that.. Note that you can't perform any " screen clicks" at this moment... 
This mode has been added for those users who would like to view YouTube or any video inside Termux's environment...


> Ps last artifact ist increased from 7Mb to 10 Mb but not suitable for further EK-X-checks?

That's because termux-shared library is added.

RalfWerner:
> Note that you can't perform any " screen clicks" at this moment...

With _Taskbar App_ it is already possible - but with the restrictions described above and of course on my Samsung device.
> That's because termux-shared library is added.

I did not understand. Screen Size for **T:X11** and presentation of the EK-Line initially have nothing to do with the (shared) symbols in it! The _show EK-X Choice_ is now missing in the **Preferences** and the top range (formerly _fullscreen choice_) is still unavailable.

suhan-paradkar:
> I did not understand. Screen Size for T:X11 and presentation of the EK-Line initially have nothing to do with the (shared) symbols in it! 

I am not speaking on C++ Shared libraries.. I mentioned `termix-shared` jitpack library

RalfWerner:
I'm aware! I meant the missing **EK-X** Line, which is initially a graphic area between Nav-Bar _or_ non-floating KB.
A positive check is the height reduction of X11 when non-floating-KB is turned on. In between should be EK, as well as in Termux.

https://user-images.githubusercontent.com/45426380/131357289-ee08154b-c80b-4ac4-b192-5b9f809df6fb.mp4

For illustration, a small video of my SM-P610 in which both Apps _in Picture_ can be seen and the OS-KB is turned on twice.
At the end, I move the BT mouse from **X11** to **Termux** and to the KB. It is well recognizable (finger cursor) _where_ the EK-T is arranged and how the picture/screen size changes with the OS-KB. If EK-X is positioned _as well as_ EK-T, there are no more coverage problems.

twaik:
About PIP mode: I think it is not applicable in all cases. Maybe we can implement mode like this one:

https://user-images.githubusercontent.com/9674930/131396770-b8776497-3f0b-4ad6-9558-e6ff3eda44e8.mp4

I do not think it will require a lot of work.

agnostic-apollo:
>That's because termux-shared library is added.

'tis not!

x11-usage1
x11-usage2

suhan-paradkar:
@twaik 
> About PIP mode: I think it is not applicable in all cases. Maybe we can implement mode like this one:

I'll consider that... Will try once this Pull Request is merged... 

@agnostic-apollo 

> 'tis not!

How?? Your Image clearly shows it... `com/termux/shared`

agnostic-apollo:
> Your Image clearly shows it... com/termux/shared

It is, but its only using `713KB` after decompilation, not `3MB` in-APK.

suhan-paradkar:
Yes.. the rest are other depenencies...

suhan-paradkar:
I think I will be adding ViewPager here itself instead of messing up the view layouts

RalfWerner:
> About PIP mode: I think it is not applicable in all cases.

Moin @twaik! It's about Extra Keys (**EK** line) here and not [Launcher](https://github.com/termux/termux-x11/pull/19#issuecomment-908642057).
The term _"PiP"_ is ambiguous. In the context of **X11**, it is also a choice that _I do **not** mean_ here, but an extension of the Split function offered on most devices. On small devices Split is the best solution to use parallel on _one_ screen with **X11** XClients while tracking with **Termux** logs, when only the phone is available.

Currently **EK** is completely disappeared in **X11**. In your original, it is still there at the end of the [Videos](https://user-images.githubusercontent.com/45426380/125047883-b241a300-e09f-11eb-9952-09bd1f22f905.mp4). Not all devices have a comparable _"in Picture"_ administration (theoretically all Android 11 devices should have it) and [Taskbar](https://github.com/termux/termux-x11/pull/19#issuecomment-908235805) is _not a solution._

The _one_ OS-KB is used jointly by _both_ apps So this should also apply to the extension **EK**. In the [Clip](https://github.com/termux/termux-x11/pull/19#issuecomment-908412984) above, only the **Termux EK** is included. If visible and both apps use the _one_ **EK** (on the right side) it is sufficient. [Here](https://user-images.githubusercontent.com/45426380/131357289-ee08154b-c80b-4ac4-b192-5b9f809df6fb.mp4) the first eight keys are visible.

But this only applies if the two apps are arranged next to each other or in Split mode possibly one above each other. Also the cases X11 _inside_ Termux are to be considered and vice versa (also in the [Videos](https://user-images.githubusercontent.com/45426380/125047883-b241a300-e09f-11eb-9952-09bd1f22f905.mp4)) and the "overlapping pictures case".

Essential is also the **X11** _only_ case. At the same **EK** as in **Termux** at the end of the **EK** is `KEYBOARD` and App Toggle  (`X11` or `TUX`) would contain. Then the _Back key_ (Nav Bar) would _not be misused_ for different functions in the two Apps.

suhan-paradkar:
Looks like the views have trouble again.... Any ideas @agnostic-apollo ??

RalfWerner:
On my slow device appears very short _"something"_ between the lower edge of the X11 screens and the lower edge of the phone screen or the Nav bar. What happens when you move the X11 screen up by 30 dpi?

agnostic-apollo:
>Any ideas @agnostic-apollo ??

I'll look into it. May take time. Busy with lot of things.

suhan-paradkar:
Well.. most of the issues with views have been fixed (not all) ... And thre are a few issues here...

- [ ] Arrow keys and control chars don't work
- [ ] Fix crashes when drawer button is pressed.. Currently, it changes with a NullPointerException and is coded to open `Preferences`
- [ ] PIP mode is partially broken and results in NullPointerException sometimes
- [ ] Extra-keys overlap the SurfaceView when app is opened but it is solved when user switches to PIP mode and comes back

RalfWerner:
### Congratulation You are a genius!
> Fix crashes when drawer button is pressed.. Currently, it changes with a NullPointerException and is coded to open

Do you've used the last **Termux** commits/artifact (known bug in an intermediate version)
> Extra-keys overlap the SurfaceView when app is opened but it is solved when user switches to PIP mode

Take a moment to celebrate - you're on a good way :) Your last artifacts results on my Gigaset following screen:
grafik

At the top right, the _limited_ [Taskbar App](https://github.com/termux/termux-x11/pull/19#issuecomment-908235805), in the middle **Termux** _in Picture_ with floating OS-KB below Android-NavBar and **EK** (identical to **Termux**) of **X11**, which fills the rest of the screens. In **X11** startet the XClients `check` and `fluxbox`  with his own taskbar, covered by **EK** (unreachable).
I mean, when EK of **X11** ist on (choice in Preferences) the X11-hight has to reduced by EK-hight, like in **Termux**.

suhan-paradkar:
Well... Something is really fishy here.... The control chars and the arrow keys don't work at all...

agnostic-apollo:
> The control chars and the arrow keys don't work at all...

In `termux-app's` case, arrow keys and other special key codes are converted to escape sequences that the terminal can understand by [`KeyHandler`](https://github.com/termux/termux-app/blob/v0.117/terminal-emulator/src/main/java/com/termux/terminal/KeyHandler.java#L156) based on what modifier keys (`ctrl`, `alt`, `shift`) are pressed.

In lorie's case there is no such handling for `ctrl` and `alt` as far as I can see and only mapping for `shift`. As for why `KEYCODE_DPAD_UP(19)` doesn't work, it would depend on mapping and who ever receives the key is handling it properly or not. Maybe @twaik knows better or the logs.

https://github.com/termux/termux-x11/blob/d78d8778c9708aa1a7a4bd73ab152f07b82ed4c2/app/src/main/java/com/termux/x11/LorieService.java#L312

https://github.com/termux/termux-x11/blob/d78d8778c9708aa1a7a4bd73ab152f07b82ed4c2/app/src/main/jni/lorie/backend/android/android-app.cpp#L278

https://github.com/termux/termux-x11/blob/d78d8778c9708aa1a7a4bd73ab152f07b82ed4c2/app/src/main/jni/lorie/backend/android/keymaps.h#L79

RalfWerner:
I think it should be soluble. In the original _worked **EK-X**_  too, so a distinction is necessary, which should be processed in the original way. At the `KEYBOARD` Key, the **X11** size should also be corrected on my phones, so that EK line and Fluxbox taskbar are _**not covered.**_ Transferril I have the Hachers-KB-Toggle as a possibility but that's not floating.

suhan-paradkar:
@agnostic-apollo Looks like those are defined here
https://github.com/termux/termux-x11/blob/d78d8778c9708aa1a7a4bd73ab152f07b82ed4c2/app/src/main/jni/lorie/backend/android/locale/android-keycodes.h#L66

RalfWerner:
In my [OTS](https://github.com/WMCB-Tech/termux-wayland/discussions/19#discussion-3476183) description is the EK example I used in my shots above. This EK (`*Properties`) is the result of an EK-Toggle function (`u ky`), which changes between 1/2-line EK and completed with `termux-reload-settings`. Unfortunately, EK-X remains unchanged until **X11** is restarted. This also applies to the EK-Choice in Preferences (corresponds to the `KEYBOARD` Toggle in the `DRAWER`)
> I have planned to make it possible to customize extra keys through the preferences of Lorie.

@twaik I can understand your wich to be independent! But the required EK of the two apps are predominantly _identical_ and where they already work like `KEYBOARD` Toggle needs nothing to be done (if EK is not covered in **X11**).
Furthermore, a lot of work was invested in the user-defined description and many users know that well.
_**Why should not that be used?**_ My original idea is just _one_ EK line if both Apps are at the same time on the [screen](https://user-images.githubusercontent.com/45426380/125047883-b241a300-e09f-11eb-9952-09bd1f22f905.mp4)

suhan-paradkar:
Well.. the control chars work now.. but looks like we hit another bug...

https://user-images.githubusercontent.com/76830230/132081536-da0dd399-9160-4a56-86af-bb2ac540a48b.mp4

As you can see above, the control chars give out continuously repeated output, instead of just only once...

agnostic-apollo:
In `termux-app`, when `TerminalExtraKeys` [calls `mTerminalView.onKeyDown()` and  `mTerminalView.inputCodePoint()`](https://github.com/termux/termux-app/blob/e7d06aebb59948730f32a2a251eb61f6a36281ad/termux-shared/src/main/java/com/termux/shared/terminal/io/TerminalExtraKeys.java#L63), the [`onKeyDown()`](https://github.com/termux/termux-app/blob/e7d06aebb59948730f32a2a251eb61f6a36281ad/terminal-view/src/main/java/com/termux/view/TerminalView.java#L703) and [`inputCodePoint()`](https://github.com/termux/termux-app/blob/e7d06aebb59948730f32a2a251eb61f6a36281ad/terminal-view/src/main/java/com/termux/view/TerminalView.java#L771) both [call the `TermuxTerminalViewClient`](https://github.com/termux/termux-app/blob/e7d06aebb59948730f32a2a251eb61f6a36281ad/app/src/main/java/com/termux/app/terminal/TermuxTerminalViewClient.java#L295) to read the control key values. They will [automatically get deactivated](https://github.com/termux/termux-app/blob/e7d06aebb59948730f32a2a251eb61f6a36281ad/termux-shared/src/main/java/com/termux/shared/terminal/io/extrakeys/ExtraKeysView.java#L615) if user had not locked the keys with long press so that next time those functions get called, they are not active. You need to create your own logic for this.

twaik:
https://github.com/termux/termux-x11/blob/e951592bbeaf339ddbf27d3eea931c62ddf54e1b/app/src/main/java/com/termux/x11/TerminalExtraKeys.java#L84

onTerminalExtraKeyButtonClick: Click means KeyEvent.ACTION_DOWN with following KeyEvent.ACTION_UP.

RalfWerner:
> As you can see above, the control chars give out continuously repeated output, instead of just only once...

I suspect that based on [Autorepeat](https://github.com/termux/termux-app/pull/2237#issuecomment-910703499), (press / release are two events) what can lead to conflicts at termux. I discover with `D` will open Preferences but no longer the session table and changes are still active after exit and restart. My checks with your EK:
- `KEYBOARD` open OS-KB between the first up to 3rd tap (same in **Termux**). non-floating-KB reduce **X11**-screen _not_, as well EK overlaps it (AVD). 
- `END,HOME` works well as expected as well as `{key:'/',popup:{display:'\\\\',macro:BACKSLASH}},{key:  '-',popup:'|'}` otherwise the Cursers, `TAB,BACKSPACE,ENTER` etc uses _autorepeat_ and stop with other Keys or continue with _autorepeat_ of this.

Currently I check the **T:Float** feature/plugin, after Taskbar check was rather frustrating. For AVD emulators this is also difficult (_pinch_ after _long tap_). Do you have experience with these @suhan-paradkar?

twaik:
@suhan-paradkar again you forgot to make additional call with KeyEvent.ACTION_UP to prevent key repeating.

suhan-paradkar:
Is [this](https://github.com/termux/termux-x11/pull/19/commits/848f55fcdd26f8f12f5fc2f053d4876ef85e9292) right or.. are there any more further improvements required? @twaik

twaik:
@suhan-paradkar Looks good. I think in a few days I will have a demo program which will show how to use T:X11 without resigning termux-app. For now I can not test it.

suhan-paradkar:
Well.. doing some tests revealed that arrow keys are good amd working fine, but `CTRL` and `ALT` are now messed up... Maybe there's some mistake.. lemme go through it....

twaik:
@suhan-paradkar These keys are not handled. Shift is the only key handled because Android did not send KeyEvent.ACTION_DOWN for shift key when I typed non English characters. 
To resolve that you should send KeyEvent.ACTION_DOWN with appropriate keyCode on "hold" click on a button and send KeyEvent.ACTION_UP on "release" click instead of  saving button state in special class.

https://github.com/termux/termux-x11/blob/848f55fcdd26f8f12f5fc2f053d4876ef85e9292/app/src/main/java/com/termux/x11/TerminalExtraKeys.java#L52-L67

twaik:
@suhan-paradkar Also I think you should make ctrl/shift/alt behaviour a bit more complicated. I mean short tap should turn button on only for one key tap, long tap should keep button on until it is switched off. And a short tap again to switch it off no matter what state it is.

suhan-paradkar:
Termux app already has the exact same behaviour.... Its just that I need to integrate this app with `termux-shared` library more effectively

RalfWerner:
@twaik That is already the behavior in **Termux** and in my check it works correct in **X11**! Only **X11**-Screen adaption and EK and _switch on/off_ is incorrect on my devices.

twaik:
As I said before I have no ability to test up-to-date build. In a few days I will try to send demo showing how to make it possible to use it without sharing the same key with termux-app.

RalfWerner:
@suhan-paradkar I would have expected the Xfce Desktop (**X11**-screen) _above_ the EK in your [Video](https://github.com/termux/termux-x11/pull/19#issuecomment-912901950) and your actions should also be "visible", even if you move aterm to the _bottom_ of the desktop. There is also the Xfce Taskbar.
This should also be operated _with EK+OS-KB_. At VNC and on my SM-P610 that works.

What happens, if you split the screen on your Oppo A12 with the _three-finger sweep_, display **Termux** below, and tap `KEYBOARD`?
Do you've also _**two EK and different behavior**_, with `KEYBOARD`?

suhan-paradkar:
Well... The `CTRL` and `ALT` keys don't work no matter what I do... am I doing something wrong here?

RalfWerner:
@suhan-paradkar Have you _**checked**_ [this](https://github.com/termux/termux-x11/pull/19#issuecomment-913093383) behavior and is your _autorepeat **bug fix**_ lost in the last Artifact?

RalfWerner:
### The _autorepeat_ Bug is fixed but not everyone else.
My expectation would be the same _key events_ as in [Hackers KB](https://github.com/termux/termux-x11/pull/19#issuecomment-913362626). But I get three times `Key code:244 state:0  Escape` only.
The `SHIFT,ALT,CTRL` keys of the OS-KB should also be considered in the _ekeys_ and vice versa (are currently independent).

But there is also good news: The `PASTE` (popup of `TAB`) translates the **Termux** clipboard buffer (`BG-(null):2000x450`=**X11** size without EK+OS-KB) in correct _key events_. Here a `check` shot in which the `END` key removes the text behind XPM in line 1 and then converts the 18 char of clipboard into 30 _key events:_
grafik
Here are the log of first five _key events_. This allows all _normal_ copy/paste actions that are implemented in most XClients (e.g. ^c/^v) are also used _normal_ **additionally to clipboard**, and an issue is also fixed, which was unfortunately also deleted with [Wayland](https://github.com/WMCB-Tech/termux-x11/discussions/4#discussion-3557430).
The **Termux** EK is toggled by [u ky](https://github.com/termux/termux-float/issues/37#issuecomment-916777123) but **X11** EK is not updated, even **not** with change of Preferences (`DRAWER` popup on `KEYBOARD`) and at all, _one_ EK would be enough!

suhan-paradkar:
Well.. @agnostic-apollo looks like I have hit a Dead end....
The bugs with arrow keys repeating themselves have been fixed tho...
The only problem is with `CTRL` and `ALT` keys.. which aren't responding at all...
Maybe you have some ideas...
If you're busy, nvm...

RalfWerner:
@suhan-paradkar The problems/bugs did **not exist** in the _X11-Init version_ yet. As well as the fullscreen mode. So that was already solved and only has to be reset. When I find time I will create a check _video._

agnostic-apollo:
Modification needs to made to `ExtraKeysView` [here](https://github.com/termux/termux-app/blob/e7d06aebb59948730f32a2a251eb61f6a36281ad/termux-shared/src/main/java/com/termux/shared/terminal/io/extrakeys/ExtraKeysView.java#L486) to send events for special buttons too instead of just maintaining state and let `lorie` maintain the state itself as mentioned by twaik [here](https://github.com/termux/termux-x11/pull/19#issuecomment-912969443) as well. The `readExtraKeysSpecialButton()` design will not work for `lorie`, since its expects separate events to be passed for special buttons key down and key up instead of a single event for a normal key + the current `metaState`/special buttons states. Only shift is handled [here](https://github.com/termux/termux-x11/blob/50c1084fee6a4b23258567f6a84f20c406d9bb57/app/src/main/java/com/termux/x11/LorieService.java#L341).

And currently, I'm not sure what you are trying to do [here](https://github.com/termux/termux-x11/blob/50c1084fee6a4b23258567f6a84f20c406d9bb57/app/src/main/java/com/termux/x11/TerminalExtraKeys.java#L162). You are reading the state into nothing. And possibly a typo of using bitwise OR [here](https://github.com/termux/termux-x11/blob/50c1084fee6a4b23258567f6a84f20c406d9bb57/app/src/main/java/com/termux/x11/TerminalExtraKeys.java#L98).

The current code will of course not work. The macros with special buttons will also not work with the new design since it uses a single event, possibly separate events could be generated but key down and key up would be problematic, like when do you send key up for `CTRL c` macro, considering if you send it at the end after `c`, then it would unset the internal state of `lorie` for `ctrl` even though a user may still have held down a dedicated `ctrl` key down elsewhere. There will likely be other issues related to this, which I can't investigate currently and you guys will have to. It would be better if `lorie` was modified to termux design if you want consistent experience. If you want the `ExtraKeysView` modifications, I can try taking time out to do that.

RalfWerner:
> It would be better if `lorie` was modified to termux design if you want consistent experience. If you want the ExtraKeysView modifications, **I can try taking time out to do that.**

@agnostic-apollo I would be very much for that. Preferably, if only _one_ **EK** is displayed for all three apps, such as OS-KB _and_ [FM=Fullscreen mode](https://github.com/termux/termux-app/pull/1923#issuecomment-918252727). This should then apply to **Termux** _and_ **Float (T&F)**. Many of the current problems with the **X11-EK** would probably be solved by _one_ **EK**.

@suhan-paradkar Thanks for the [Bug Fix](https://github.com/termux/termux-x11/pull/19#issuecomment-922604383) and the **RS=Reseed Screen While Soft Keyboard is Open** choice extension (_Configure Events_). There are basically and opposite to init-**X11** still some problems open, so I do not know which version is better. Are you still working on it and at ekeys->master branch merge?

Meanwhile, I found a [AVD](https://github.com/termux/termux-x11/pull/19#issuecomment-907089567) in which _no flickers_ occur. Here seven Shots from this `Galaxy_Nexus_API_30` to show the effects of **FM. T&F** just do _not_ have them. The last two above show the **FM** and the lower reduced **X11** configurations. Example after [OTS](https://github.com/termux/termux-float/issues/37#issuecomment-916777123):
- 1 notifications for the started apps **X11, T&F**. In this screen, **Float** will _not_ be displayed and Preferences can be selected.
- 2 init-**X11** Preferences with **Float** superposition. Here **FM** can be set (missing in the new-**X11**) but **RS** is missing. **Float** is transparently displayed after OS-KB removed with back-key and moved with _long tap_.
- 3 **X11** uses the entire screen. The two NavBars from the shot2 would be briefly displayed with sweep from above or below, such as [here](https://user-images.githubusercontent.com/45426380/133095986-50fcd2b4-8b2a-4e33-97e7-1977a5197522.png), the two `touch` windows and **Float** are completely displaceable throughout the **BG**=wallpaper/background (also managed by XClient `touch`).
- 4 like Shot3 but in rotated to Landscape and _non-floating_ OS-KB to **X11** (back-key). This can only be used in the area restricted by **Float**. `touch` was started in **Float** and the X-events logged there.

grafik
- 5 like shot4 but after _tap_ on **Float**. Key events are now being processed in **Float**, but not visible because the **Float** Picture is too high for the rest screen without NavBar and OS-KB. The reduction of the **Float** size (_long tap+pinch_) is very difficult on AVD because only the mouse can be used. Here it would also fail because there is a minimum height for float. To my surprise, the X11 EK is displayed here, which I expected in Shot4, but is meaningless.
- 6 like shot5 but with vertical split and termux right. This AVD allows only split in the middle (horizontal split line). Both indicated EK can _**not**_ be used here in **Float**. Since **RS** is not included in init-**X11**. For example, no X11-configure events are made, and then `touch` is unable to position the relevant areas in the visible range. This also applies to Shot4+5
- 7 like Shot6, rotateed in Portrait (Vertical Split line), Iconization of **Float** and **Termux** with `KEYBOARD`. Here, only EK remains from the **X11** screen and _no X-events_ are possible. Conversely, it is not better, which is why I did not shot any of it.

Summary: _**one**_ EK and **FM+RS** as Default would be the solution. For all three I do not need a _choice_ in X11 Preferences and **Float** could be effective used. The other Preferences choices also does not have benefit me, so that only **Touchscreen input** selection remains and this could be displayed directly.

What is not treated here concerns the EK/OS-KB _Feetback_. I still work on a meaningful video...
But this is not **X11** specific, as it also affects **T&F** and I would use Hackers-KB (more toggle keys) instead of [SM-KB](https://user-images.githubusercontent.com/45426380/131357289-ee08154b-c80b-4ac4-b192-5b9f809df6fb.mp4).

suhan-paradkar:
> It would be better if lorie was modified to termux design if you want consistent experience. 

@agnostic-apollo well.. I guess then I will have to modify lorie then... But I'm not sure what modifications need to be done though

RalfWerner:
@suhan-paradkar The _key->state_ is missing see [above ff](https://github.com/termux/termux-x11/pull/19#issuecomment-913362626) (was correct in _init-**X11**_) and branch merge _ekeys->master_

suhan-paradkar:
Well.. I've modified Lorie a bit here... But it doesn't seem to work as of now... I wonder what else need sto be done🤔🤔🤔

RalfWerner:
Nice to see you active here. In which state is your current APK, is it _**suitable for check?**_

After the decoupling (run-command) of the Ekey-Branch no acquisition of the EK from `~/.termux/t*`, as before and the installation process is more error-prone and confusing`*)`. The _Fullscreen_ option removes the small upper bar but _**not**_ the larger below/right as the `*VNC` APK.
I had described [here](https://github.com/bk138/multivnc/issues/161#issuecomment-988107258) checks of four `*VNC`. Compared to the version of yesterday, the EK can also be seen in split mode:) The _Key state_ (Shift, Ctrl, Old) are _**still not set. What do I have to do so that EK is taken over?**_

`*)` Previously, only the _Xwayland_ package was installed and restarting the **X11**+Xwayland for updating the EK or start **PID4**=XClients (quite simple). Now, the additional `*.deb` has to be installed locally, which exists in an older version also in the X11 repo. In addition, `termux-startx11` processes (is superfluous for **PID3**) are started/closed several times (quite complicated).

suhan-paradkar:
Well.. Full-screen workaround is quite buggy as of now and I intend to fix it.. as on Ctrl and Alt keys... I'm still looking on options

suhan-paradkar:
Well.. looks like Termux:X11 Isn't able to read Shared properties... It loads the default extra-keys instead of the ones from `termux.properties`. Do I need to change something @agnostic-apollo ?

RalfWerner:
EK in the current `master` branch is very limited (8 keys without `popup` and always fixed as in the 2 year old `x11-basic`). But _Key-State_ is correctly implemented here in X events - maybe it help you!

The **PID3**=Xwayland process (for **X11** App) of Vers: `x11_02_06`, has to start with two delays (sleep) if it does not exist yet. That was much easier at Vers: `x11_02.04` (_one_ sleep after the App Start and t`ermux-x11.deb` is superfluous) and the EK-Properties were also read. **_Can this possibly be reset?_**
PIP should also be _off_ at the first start after installation

### Another feature request to Extra-Keys lines:
EK in Termux are in `~/.termux/*.properies` edited but different or not implemented by the Apps: **Termux, Float, X11** and possibly **MultiVNC**. After each change, EK should be updated in the Apps.
A treatment _"like OS-KB=OnScreen KeyBoard"_ would be optimal and means: only _**one**_ EK, for _all Apps_. Display as floating or non-floating and at non-floating always outside (below) of the RS=Reseed Screens and with OS-KB above it (at Termux, that's the state - only currently without fullscreen). Use with FullScreen, Split/Popup View should completely display the EK. Preferences for FullScreen and RS=Reseed Screen are then no longer used.

The current description options would have to be extended by a few keys for **X11**. First, the Mouse Buttons if you are to be emulated by finger actions. Example/suggestion with 11 Keys in _one_ line:
```
#extra-keys=[[{key:MB1,popup:DRAWER},{key:MB2,popup:KEYBOARD},{key:MB3,popup:PASTE},{key:SHIFT,popup:ENTER},{key:CTRL,display:'^',popup:ALT},{key:UP,popup:PGUP},{key:DOWN,popup:PGDN},{key:LEFT,popup:BKSP},{key:RIGHT,popup:TAB},{key:END,popup:HOME},{display:X11,macro:'t SPACE 0 ENTER',popup:'Termux'}]]
 extra-keys=[[{key:DRAWER,popup:MB1},{key:KEYBOARD,popup:MB2},{key:PASTE,popup:MB3},{key:ENTER,popup:SHIFT},{key:CTRL,display:'^',popup:ALT},{key:UP,popup:PGUP},{key:DOWN,popup:PGDN},{key:LEFT,popup:BKSP},{key:RIGHT,popup:TAB},{key:END,popup:HOME},{display:X11,macro:'t SPACE 0 ENTER',popup:'Termux'}]]
```
Some Keys could be fixed with Long Tap and contain `popup` Keys that are sometimes reversed in the `#`-line. Fixing in `popup` is currently not working. _**Can you repeat this with Vers: `x11_02.04`?**_ In **X11* (+**float**) and _portrait_ it looks like:
grafik
and in **Termux** (+**float** icon) _landscape_ and revers Keys it looks like:
grafik

agnostic-apollo:
> It loads the default extra-keys instead of the ones from `termux.properties`. Do I need to change something @agnostic-apollo ?

The `master` branch does not use `sharedUserId=com.termux` anymore. You can't access termux files any more. f9a9ce31

After requesting `com.termux.permission.RUN_COMMAND` in `AndroidManifest.xml`

`<uses-permission android:name="com.termux.permission.RUN_COMMAND"  />`

 you should be able to add following to `TermuxAppSharedProperties`.  Preferably `runTermuxContentProviderReadCommand` should be in a different util class. I may add it to `termux-shared`.

```

private TermuxAppSharedProperties(@NonNull Context context, File propertiesFile) {
    super(context, TERMUX_X11_APP_NAME, propertiesFile,
    TermuxPropertyConstants.TERMUX_PROPERTIES_LIST, new SharedPropertiesParserClient());
}

public static TermuxAppSharedProperties build(Context context) {
    Uri propertiesFileUri = Uri.parse("content://" + TermuxConstants.TERMUX_FILE_SHARE_URI_AUTHORITY + TermuxConstants.TERMUX_PROPERTIES_PRIMARY_FILE_PATH);
    File propertiesFile = new File(context.getFilesDir(), "termux.properties");
    runTermuxContentProviderReadCommand(context, propertiesFileUri, propertiesFile);
    return new TermuxAppSharedProperties(context, propertiesFile);
}

private void runTermuxContentProviderReadCommand(Context context, Uri inputUri, File outFile) {
    InputStream inputStream = null;
    FileOutputStream fileOutputStream = null;

    try {
        inputStream = context.getContentResolver().openInputStream(inputUri);

        if (!outFile.exists())
            outFile.createNewFile();
        fileOutputStream = new FileOutputStream(outFile);
        byte[] buffer = new byte[4096];
        int readBytes;
        while ((readBytes = inputStream.read(buffer)) > 0) {
            // Log.d(LOG_TAG, "data: " + new String(buffer, 0, readBytes, Charset.defaultCharset()));
            fileOutputStream.write(buffer, 0, readBytes);
        }
    } catch (Exception e) {
        e.printStackTrace();
    } finally {
        try {
            if (inputStream != null)
                inputStream.close();
            if (fileOutputStream != null)
                fileOutputStream.close();
        } catch (IOException e) {
            e.printStackTrace();
        }
    }
}
```

And then call it like following in `MainActivity`. 

```
protected void onCreate(Bundle savedInstanceState) {
    ...
    mProperties = TermuxAppSharedProperties.build(this);
   ....
}
```
I haven't tested this. You will need latest github debug build for `termux-app` or `v0.118` (when released) for this to work. You may also want to think on whether you actually want to share properties between both apps with new design.

suhan-paradkar:
> You may also want to think on whether you actually want to share properties between both apps with new design.

Well.. I prefer it since it gives a sense of continuity between both apps....

> you should be able to add following to TermuxAppSharedProperties. Preferably runTermuxContentProviderReadCommand should be in a different util class. I may add it to termux-shared.

I will try...

suhan-paradkar:
Well.. it's fixed now.... Thanks to @agnostic-apollo . I Will need to dofurther work on the CTRL and ALT keys

suhan-paradkar:
Well.... @twaik looks like Extra-keys works without resigning either of the apps...

twaik:
It works if you have debug version. But it will not work with latest Termux release. That means we need to fix jni in termux too.

RalfWerner:
Nice! The acquisition of the EK works on the _**first X11 start**_ and using the latest **Temux+X11** versions.

Subsequent EK change or changing the appropriate preference: ON/OFF are _**only accepted by completely termination of X11 and restart PID2-4!**_ This process is often required for check. Can _**X11 only be stopped with the notification?**_

The correct RS function from **X11_02.04** has unfortunately been _**lost again!**_ OS-KB covers the EK and also the entire lower area of ​​the RS (see [above](https://github.com/termux/termux-x11/pull/19#issuecomment-997229208)). @agnostic-apollo had used many attempts until this properly worked in Termux on all devices - a joint EK would be better.

The complicated `termux-x11` Start is unfortunately even more complicated and error-prone as I thought. OTS=PID1-4 expects **PID3** before **PID4** (nice and easy at **X11_02.04**). Here, however, the sleep phase between the two processes is to choose longer, as **Xwayland** exists although it is not usable immediately. Also, the "Zombie-**PID4**" processes (**PID3** is active and **PID2** stopped) occur more frequently. I am uncertain whether the _"Run Command"_ effort justifies the benefit.
_**Do you use X11 with other Apps as Termux?**_

Not only _CRTL,ALT_ (Key-code: 4,8) but also _Shift_ and possibly _Start_ (Key-code: 1, 64) as with ***VNC** should work. If the _"OS-KB covered the EK"_ problem is solved, Hackers-KB can easily be compared the correct implementation. (see [here](https://github.com/termux/termux-x11/pull/19#issuecomment-913362626) and [here](https://github.com/termux/termux-x11/pull/19#issuecomment-921708037))

suhan-paradkar:
@agnostic-apollo can you please make `onAnyExtraButtonClick` function [here](https://github.com/termux/termux-app/blob/e7d06aebb59948730f32a2a251eb61f6a36281ad/termux-shared/src/main/java/com/termux/shared/terminal/io/extrakeys/ExtraKeysView.java#L485) public, instead of private ?

RalfWerner:
@suhan-paradkar I had asked you above a few questions. _**Can you please give a short (yes/no/in progress) answer?**_

One more: [OTS](https://github.com/termux/termux-float/issues/37#issuecomment-916777123)=OneTouchStart with **VNC** is more stable than with **X11** because PID2 start can be retrofitted (also different [Viewer Apps](https://user-images.githubusercontent.com/45426380/138826536-8a696f46-e187-488d-a452-8c6e0a2e4433.mp4) simultaneously), without repeating the PID3=Xvnc + PID4 start. _**Could the behavior**_ (do _not_ kill PID3=Xwayland with PID2) at **X11** are _**also achieved?**_

suhan-paradkar:
> The Key state (Shift, Ctrl, ALT) are still not set. What do I have to do so that EK is taken over?

 Its still a work in progress

> In addition, termux-startx11 processes (is superfluous for PID3) are started/closed several times (quite complicated).

Yes.. I've noticed.. the init process is quite buggy

> Can X11 only be stopped with the notification?

As of now.. yes...

> Do you use X11 with other Apps as Termux?

Actually.. right now, we can't... The goal is to support 3rd party apps to support x11 output through this app...

> Could the behavior (do not kill PID3=Xwayland with PID2) at X11 are also achieved?

I guess we can....

agnostic-apollo:
> @agnostic-apollo can you please make `onAnyExtraButtonClick` function public, instead of private ?

For now, use `implementation 'com.github.agnostic-apollo.termux-app:termux-shared:temp-extrakeys-SNAPSHOT'`

https://github.com/agnostic-apollo/termux-app/tree/temp-extrakeys

Will push it to `master` later.

RalfWerner:
> Its still a work in progress ... I guess we can....

These are good news :) I hope the next version for check contains again the lost functions: **RS** (`02.04`), **Key-State** (`master/basic`) and **FullScreen** (`basic`). In addition to `touch` and `check` I also use **` multi`**, which often crashes or remains as a zombie process. This is a C source of the XInput2 developer and uses exclusively `Direct mode` for fingers and S-Pen events.

agnostic-apollo:
@suhan-paradkar  What is your end goal here?

The `TerminalExtraKeys` does not extend from `ExtraKeysView` so its not a child class, it only implements its interface `IExtraKeysView`. If you want to call a function of `ExtraKeysView`, pass its object to `TerminalExtraKeys`  constructor and then call it via the object, it "should" work. For `onAnyExtraKeyButtonClick()`, you can just pass `ExtraKeyButton`, the `View` and `Button` can be `null`, as long as you don't use them in interface implementation of `onExtraKeyButtonClick()`.

https://github.com/termux/termux-x11/pull/19/files#diff-611e4439e114c1564ad8ad5fde76e48586f3f5bf2cec31fb5ff86552aa7fd2bbR48

suhan-paradkar:
> The TerminalExtraKeys does not extend from ExtraKeysView so its not a child class, it only implements its interface IExtraKeysView. If you want to call a function of ExtraKeysView, pass its object to TerminalExtraKeys constructor and then call it via the object, it "should" work

@agnostic-apollo You mean something like this?

```
public TerminalExtraKeys(@NonNull View.OnKeyListener eventlistener, MainActivity mact, ExtrakeysView extrakeysview) {
        mEventListener = eventlistener;
        act = mact;
        extrakeysview = mextrakeysview;
    }
```

agnostic-apollo:
> `extrakeysview = mextrakeysview;`

Typo

`mExtraKeysView = extraKeysView`

You should also use camel case convention. And non-public class instance variables should ideally start with `m`.

https://source.android.com/setup/contribute/code-style#follow-field-naming-conventions

suhan-paradkar:
Closing as mentioned in https://github.com/termux/termux-x11/discussions/46#discussioncomment-3089408

Issue #21, opened 2021-08-27
[Feature request] Using termux-wayland without need to compile it with Termux key.


twaik:
Hi guys! I have no time to help to the project but I have an idea how to use termux wayland without signing it with the key Termux::App is signed. 
The idea is to use the same connection mech Termux::API uses. Look:
1. Small program in Termux fs creates creates connection to Java service the same way Termux API package does (look here: https://github.com/termux/termux-api-package/blob/master/termux-api.c)
2. The same program creates Xserver well known sockets and sends fds to Java service. 
3. Java service sends these fds to xserver and xserver can make SetNotifyFd to these sockets (look here: https://github.com/freedesktop/xorg-xserver/blob/18d3131f9a332096825c09106a931c427246ddb7/os/connection.c#L282 )
It will let us use Unix sockets even if application has another keys. And it is compliant with SELinux security restrictions. Also in the future it will let us use android-emugl modified for Termux  to let us use fast GLES (it can be modified to use Unix socket). What do you think?

twaik:
Sorry, my fault. I forgot that T::X11 uses Xorg on Termux fs side. We can try to provide wayland socket connection through the Java service.

twaik:
After receiving socket fd in Java service you can push it to wayland server using wl_display_add_socket_fd function on JNI side.

twaik:
Ok. I was not right. It is impossible to send Unix socket fd the same way Termux::API does. But we can try to send [ACTION_VIEW](https://github.com/termux/termux-packages/issues/7437#issuecomment-909527512) broadcast with URI to temporary Unix socket. T::X11 will be able to open it and there we will be able to send fd to our socket. 
T::X11 needs something like [TermuxFileReceiverActivity](https://github.com/termux/termux-app/blob/e7d06aebb59948730f32a2a251eb61f6a36281ad/app/src/main/java/com/termux/filepicker/TermuxFileReceiverActivity.java) but with reaction only on specific mime type (like "termux/x11"). This way it will receive READ permission on our temporary Unix socket.
That is the simplest way to make it work with Termux without resigning it and without making any change to Termux::App.
I'll try to make a demo this weekend.

suhan-paradkar:
Maybe we can make use of this [gist](https://gist.github.com/RajithaKumara/2700e362ff231da265aee25f7aa0a57a)?

twaik:
[...]

PR #31, opened 2021-09-11
Let Termux:X11 run without sharedUserId="com.termux"

twaik:
Solves #21
Pay attention. After compiling you should copy `$repo/starter/build/intermediates/stripped_native_libs/debug/out/lib/$TARGET_ARCH/libstarter.so` (or extract from apk (`lib/$TARGET_ARCH/libstarter.so`)) to `$TERMUX_PREFIX/libexec/termux-x11/libstarter.so`. Also you should put `$repo/starter/build/outputs/apk/debug/starter-debug.apk` to `$TERMUX_PREFIX/libexec/termux-x11/starter.apk`
I think we can use [this build.sh](https://github.com/termux/termux-packages/blob/2ac2f09d886bd320847cdf2ada2a28f67c636de7/packages/termux-am/build.sh) as a reference.

P.S. Almost forgot $repo/termux-x11 script. It should be copied to $TERMUX_PREFIX/bin/

- [x] Remove/ignore `build.gradle` changing to prevent downgrading gradle plugin.
- [x] Remove hardcoding `$XDG_RUNTIME_DIR` in sources. Replace with checking if it is set in environment and export default value if it is not exported.
- [x] Start `Xwayland` if/when connection is established.
- [x] Send `RUN_COMMAND` intent if an app was started from launcher, not from commandline.
- [x] Implement a way to extract logs. Maybe we will pass `fd` of logfile next to socket's `fd` and write output of logcat (logcat started in an app can extract logs of the app and apps with the same `android:sharedUserId`). 
- [ ] Implement a way to read `termux.properties` file. Using `fd` sending or through `Termux`'s `ContentProvider`
- [x] [A way to avoid using JNI to extract non-basic arguments (not int/long/boolean types).](https://github.com/termux/termux-x11/pull/31#issuecomment-917684124) 
- [ ] Implement mechanism of avoiding errors if there is an API version mismatch. Maybe we can just not perform an action if there is RemoteException error.


twaik:
@xeffyr Hi. Can we make a new package for apt repos or we should test it longer?

build.gradle
```diff
@@ -7,7 +7,7 @@ buildscript {
         
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:7.0.2'
+        classpath 'com.android.tools.build:gradle:4.1.1'
```
suhan-paradkar, line 10: Any specific reason to degrade version? I mean.. I don't prefer to degrade versions of depenencies unless it causes a problem

starter/src/main/jni/starter.c
```diff
@@ -0,0 +1,110 @@
...
+#include <sys/poll.h>
+#include <sys/un.h>
+#include <errno.h>
+#include <string.h>
+#include <fcntl.h>
+
+#define unused __attribute__((__unused__))
+
+#define PREFIX "/data/data/com.termux/files/usr"
+#define XDG_RUNTIME_DIR PREFIX "/tmp"
```
suhan-paradkar, line 18: Also.. please don't hard-code `XDG_RUNTIME_DIR` We currently use user-defined `XDG_RUNTIME_DIR`.
You can see `app/src/main/res/xml/preferences.xml`

starter/src/main/jni/starter.c
```diff
@@ -0,0 +1,110 @@
...
+#include <sys/types.h>
+#include <sys/poll.h>
+#include <sys/un.h>
+#include <errno.h>
+#include <string.h>
+#include <fcntl.h>
+
+#define unused __attribute__((__unused__))
+
+#define PREFIX "/data/data/com.termux/files/usr"
```
suhan-paradkar, line 17: It would also be better to use user-defined `$PREFIX`.. that would add support for almost every third party apps out there...

suhan-paradkar (changes requested):
> Pay attention. After compiling you should copy $repo/starter/build/intermediates/stripped_native_libs/debug/out/lib/$TARGET_ARCH/libstarter.so (or extract from apk (lib/$TARGET_ARCH/libstarter.so)) to $TERMUX_PREFIX/libexec/termux-x11/libstarter.so. Also you should put $repo/starter/build/outputs/apk/debug/starter-debug.apk to $TERMUX_PREFIX/libexec/termux-x11/starter.apk

Abou that... I suggest using termux's own `RUN_COMMAND` intent

build.gradle
twaik: Android Studio did not want to sync a project on my PC. I'll remove this change.

starter/src/main/jni/starter.c
twaik: Ok, I'll fix it. I need to check how to read it inside starter.apk.

starter/src/main/jni/starter.c
twaik: I can pass socket's path from Java code, that is not a problem.

twaik:
>  I suggest using termux's own `RUN_COMMAND` intent

It should be done just one time after package installation. Do you think I should install it using `RUN_COMMAND`, not `pkg`/`apt`?

suhan-paradkar:
> It should be done just one time after package installation

Then I guess adding it to `termux-packages/packages/termux-x11` would be better

twaik:
`termux-x11` script is needed to run starter.apk which creates socket and sends server file descriptor to termux-x11 app. It is impossible to use termux's private directories without `android:sharedUserId`.

twaik:
Also I forgot to complete a code which starts Xwayland in the end. Maybe we will make something like starting XFCE/other DE session after Xwayland. That will allow us to send `RUN_COMMAND`  to termux to avoid writing any commands. Less movements. And also it will start all the stuff after receiving `SECONDARY_HOME` intent.

twaik:
@suhan-paradkar do you think an app should check API version used in `starter.apk`? That will help to avoid segfault when versions of `termux-x11` and `starter.apk` do not match.

suhan-paradkar:
I guess better would be to add a logger to an app and also bump the version of `termux-x11` script whenever `starter.apk` is updated..  this means a `pkg update` by the user wil solve it all

twaik:
That can be a bit problematic. I mean mechanism of extracting logs. Maybe I can add extra argument for extracting log of previous run. I think I will do that after cleaning up the code and separation debug/verbose and release logging mode.

twaik:
There is a problem. When JNI code tries to use `JNIEnv* env` it segfaults. And `JNI_OnLoad` function gets `JavaVM *vm = NULL`. What is a problem? Maybe there is a solution?
@agnostic-apollo @xeffyr

starter/src/main/jni/starter.c
twaik: Ok. The situation is simple: termux-x11 app does not handle creating sockets anymore, it is handled in starter.apk. I can put there checking if XDG_RUNTIME_DIR is set and if not export default value.

starter/src/main/java/com/termux/x11/starter/Starter.java
```diff
@@ -0,0 +1,309 @@
...
+    private native boolean checkWaylandSocket();
+    private native int openLogFD();
+    @SuppressWarnings("FieldMayBeFinal")
+    private static Handler handler;
+    static {
+        Looper.prepare();
+        handler = new Handler();
+
+        @SuppressLint("SdCardPath")
+        final String libPath = "/data/data/com.termux/files/usr/libexec/termux-x11/libstarter.so";
```
suhan-paradkar, line 299: Considering adding support for 3rd party apps, I won't considering hardcoding this path...

I would prefer using this only if `DEFAULT_PREFIX` is unset.. just like in `starter.c`

suhan-paradkar (commented):
So far so good

starter/src/main/java/com/termux/x11/starter/Starter.java
twaik: Starter is needed only to trigger a termux-x11 to start without `Context`. 3rd party apps will be able to trigger termux-x11 inside `Activity`. I will add notes about that ot Readme before release. Or at least after reviewing and before merging this pull request.

twaik:
I think it is ready to be merged. 
[Last artifact](https://github.com/termux/termux-x11/actions/runs/1295555854) contains both apk and deb file of companion package.
@suhan-paradkar @WMCB-Tech 
Can you please test it?
Maybe I should add something?

RalfWerner:
I checked your Artifact on one of my devices but not studied your descriptions above (do it tomorrow), and made comparison:
- Installation as the update of the last version is not possible (only after de-install)
- **Fullscreen mode** is not included and the **Extra Keys** correspond to your base version (not ekeys).
- As **XServer** I could _not_ use this app by [OTS](https://github.com/termux/termux-float/issues/37#issuecomment-916777123) but get short messages.

twaik:
@RalfWerner 
1. This version is signed with different key. That is a reason update process fails.
2. Fullscreen mode is in `ekeys` branch which is still not merged. There is no reason to use it as a base branch.
3. After tapping on application icon in launcher T:X11 triggers `termux-app` to run script `$PREFIX/libexec/termux-x11/termux-startx11`. Also this script may be used to start X11 session from Termux session.
This script starts `termux-x11` binary to create and export socket to `termux-x11` app, starts `Xwayland` and `x-session-manager`.
`x-session-manager` is symlink to `xfce-session`. You can override it to be a link to your script.
You do not need to use 3rd party scripts starting `termux-x11`. The process of starting it has changed and it is incompatible with previous versions.

WMCB-Tech:
> I think it is ready to be merged. 
> [Last artifact](https://github.com/termux/termux-x11/actions/runs/1295555854) contains both apk and deb file of companion package.
> @suhan-paradkar @WMCB-Tech 
> Can you please test it?
> Maybe I should add something?

Good morning! during the installation it fails for some reason. so i have to switch to Termux F-droid or something @twaik ?

WMCB-Tech:
the `termux-x11` package dependency `Xwayland` needs to be `xwayland` since X shouldn't capitalized. repacking the deb fixed it

RalfWerner:
> Ok now it works after uninstalling and reinstalling the app. felt dumb

@WMCB-Tech My [experiance](https://github.com/termux/termux-x11/pull/31#issuecomment-932469616)! Where dit you find  `$PREFIX/libexec/termux-x11/termux-startx11` to start `xwayland` process?

twaik:
> the termux-x11 package dependency needs to be xwayland and not Xwayland since it's case sensitive apt will report a non-existent package. repacking the deb fixed it

Fixed

> Where dit you find $PREFIX/libexec/termux-x11/termux-startx11 to start xwayland process?

It is present in companion package. Now installation is fixed.

> What practical advantage should have this branch? Its more worse than basic X11

It is better because it can be used with any version of termux, even signed with another key. There are not many people who rebuilds/resigns their application in order to guarantee compatibility.

twaik:
@WMCB-Tech Does it work well? Do you think it can be merged?

WMCB-Tech:
> @WMCB-Tech Does it work well? Do you think it can be merged?

I just moved to F-droid version of Termux for testing and it seems it does work fine (although additional steps need to be done like permission granting and setting `allow-external-apps = true` needs to be set)

i think it's ready to merge as it's works flawlessly not having to use debug builds of Termux

twaik:
> additional steps need to be done like permission granting

It is impossible to override it. And it is needed only to start `termux-startx11` command from external app.

> `allow-external-apps = true` needs to be set

I prefer not to do that automatically because it can impact termux-app's security. 


Ok. I am merging it.

PR #60, opened 2021-11-05
[Badge Fix] fix termux Discord server ID

SaicharanKandukuri:
found that shield badge showing read badge cause incorrect SERVER ID and watching the same from ( like ~6 days )
image

so I fixed it!

But 
image
to fix that grey badge discord admins need to enable widgets in discord server settings so that it shows the number of people online

have a nice day developing termux-x11

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 
 # Termux:X11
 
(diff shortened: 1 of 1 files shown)


WMCB-Tech:
well this also needs to be fixed in https://github.com/termux/termux-app and https://github.com/termux/termux-packages as well in their README's

PR #108, opened 2022-04-03
Update MainActivity.java FLAG.FULL_SCREEN added

ask9027:
It is required to enable full screen mode


suhan-paradkar:
I guess it's better to create a preference for the user to chose if he/she wants fullscreen...

ask9027:
> I guess it's better to create a preference for the user to chose if he/she wants fullscreen...

Maybe but it will give more viewing space to user

PR #204, opened 2023-01-31
Fixed termux-x11 hanging after closing the X11 session and enabled fullscreen as fallback.

polkaulfield:


twaik:
`svc.finish()` tells to starter `your work finished`. It should not finish any activity of app.

PR #206, opened 2023-01-31
Added a fullscreen toggle and fixed the X11 session bug.

polkaulfield:


app/src/main/java/com/termux/x11/LorieService.java
```diff
@@ -212,7 +212,7 @@ public int onStartCommand(Intent intent, int flags, int startId) {
             Log.e("LorieService", action);
             terminate();
             sleep(500);
-            act.finish();
+            act.finishAffinity();
```
twaik, line 215: `finishAffinity` is not needed. `System.exit(0);` finishes the whole process.

polkaulfield: Fixed the pr!

PR #209, opened 2023-02-02
Added a setting to force hiding the software keyboard at launch

polkaulfield:
It's useful to have the ability to disable the software keyboard. Also having my USB keyboard connected made the IME pop-up anyways.

### DIFF
--- a/app/src/main/assets/en_us.xkbmap
+++ b/app/src/main/assets/en_us.xkbmap
@@ -1,7 +1,7 @@
 xkb_keymap {
 xkb_keycodes "(unnamed)" {
 	minimum = 8;
-	maximum = 708;
+	maximum = 255;
 	<ESC>                = 9;
 	<AE01>               = 10;
 	<AE02>               = 11;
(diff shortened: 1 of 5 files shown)


twaik:
I think you are late...

twaik:
I pushed my commit 8 minutes ago :)

polkaulfield:
We were looking into the same issue at once then xD. Gonna try your patch asap.

twaik:
Can you please try to write some gui preference for resolution? Something that will let user choose between resolution presets (like 5:4 / 16:10 or something else), scale (0.1x to 2x) or exact resolution? I am not skilled enough in gui.

twaik:
So what do you think?

polkaulfield:
Added a couple menu options, just the gui. If you want i can try to implement the logic too.

twaik:
No need to. All the code related to EGL and GLES is deprecated and will be removed. I will post the code soon.

polkaulfield:
#214 I added the menus here

PR #276, opened 2023-03-15
Changes in `README.md` and `workflows`

ask9027:
same changes are made in `README.md`.
these workflow changes will `cancel` previous `runs` if there are any running.
also save build time.
`build` and `test` will ignore if changes are made in `README.md` and `img/` dir.

### DIFF
--- a/.github/workflows/debug_build.yml
+++ b/.github/workflows/debug_build.yml
@@ -1,45 +1,65 @@
-name: Build
+name: Build APK
 
 on:
   push:
     branches:
       - master
       - x11-client-experimental
+    paths-ignore:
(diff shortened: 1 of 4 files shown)


.github/workflows/debug_build.yml
```diff
@@ -1,45 +1,65 @@
...
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+
+      - name: Set up JDK 17
+        uses: actions/setup-java@v3
+        with:
+          java-version: "17"
+          distribution: "temurin"
+
```
twaik, line 41: Why do we need to setup JDK17 here??

PR #323, opened 2023-05-25
make separate artifacts

ask9027:
Users can download `termux-x11` files according to their need which no need to download all files in one file. it will make files in small size which is easy to download.


twaik:
I think shell-loader and companion package should be packed to one zip package. It does not really make sence to distribute them separately.

ask9027:
> I think shell-loader and companion package should be packed to one zip package. It does not really make sence to distribute them separately.

okay will do it.

PR #357, opened 2023-06-20
fix and add in preferences for `Keep Screen On`

ask9027:
In previews `keep screen on` not working so there I make it separate in preferences, so user can makes it enable on their demands.


twaik:
Can you make `keepscreenon` camel case?

ask9027 (commented):
fixed camel case in `keepScreenOn`

PR #358, opened 2023-06-21
Added suport for stylus with a menu to switch between mouse click lef…

Dr4kzor:
Now stylus works, stylus hover and touch work, added ability to select between left right or middle click, this does not affect any input other than stylus meaning you can still use your fingers or mouse for left click while the stylus is on right click only mode


app/src/main/java/com/termux/x11/LoriePreferences.java
```diff
@@ -142,6 +142,14 @@ void updatePreferencesLayout() {
                     Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU
                     && ContextCompat.checkSelfPermission(requireContext(), POST_NOTIFICATIONS) == PERMISSION_DENIED;
             findPreference("requestNotificationPermission").setVisible(requestNotificationPermissionVisible);
+
+
+
+            findPreference("displayScale").setVisible(true);
```
ask9027, line 148: this should remove?

Dr4kzor: Hi I fixed the commit removing that unwanted change, not sure how it got there.

parkerlreed:
Tested on Samsung Tab S8+ on Android 13

Installed main apk, shell loader apk, and the updated termux-x11-nightly deb in Termux

Launched with

```
termux-x11 :1
```

New session

```
export DISPLAY=:1
dbus-launch --exist-with-session startxfce4
```

Everything works and displays to Termux-X11 but going to the notification dropdown and hitting Preferences crashes the application.

Touch is working as trackpad and stylus is controlling the mouse directly, however the stylus does not show up as a stylus to X11. GIMP input devices only shows Core Pointer.

twaik:
> Touch is working as trackpad and stylus is controlling the mouse directly, however the stylus does not show up as a stylus to X11. GIMP input devices only shows Core Pointer.

It can not show up as as stylus to X11, his code sends mouse events to server. 

I was looking for pure stylus X11 driver (to lend some code) but did not find it. Libinput code is not clear enough to understand it, wacom is for tablets and not for stylus-only devices.

parkerlreed:
OH so this just added *back* the stylus working as a mouse with some extra preferences. Thanks for the clarification!

Excited to see if this can be expanded for pressure in the future.

twaik:
I did not implement stylus support intentionally. Removing stylus support was not intentional too.

twaik:
Why did you put stylus handling code to hardware mouse handler?

parkerlreed:
Figured out what the preferences crash was, just had to reset the application data and that is working again. 

Stylus tap does whatever the configured action is, but is not moving the mouse within the session.

parkerlreed:
Stylus input is only working with touch screen input mode set to emulated touch screen. 

Direct touch which is what I prefer for the finger, does not allow the stylus to work.

twaik:
Currently I do not have stylus-enabled device for testing. Can you please test clicking with stylus? I read the code and think that it allows to click but do not allow to move cursor in direct touch mode.

Dr4kzor:
I sadly don't have time atm to edit this, and I can confirm that the current code version does not work with direct input mode, also the menu auto closing after choosing what click type to use is not user friendly (cannot use a finger and stylus at the same time to change modes comfortably). Also my stylus btns don't work so I cannot test if these are cycling between click types.
As soon as I have time I will look into this.

twaik:
aae940caeb9afab97f42339190c9473d280b9c61 should fix cursor moving. I've implemented right button click with S Pen (you should press stylusbutton before click to make it produce right click). But I do not know if it works with other devices.

parkerlreed:
> [aae940c](https://github.com/termux/termux-x11/commit/aae940caeb9afab97f42339190c9473d280b9c61) should fix cursor moving. I've implemented right button click with S Pen (you should press stylusbutton before click to make it produce right click). But I do not know if it works with other devices.

Indeed it does! Direct touch stylus and the right click when button held is working great. Thank you.

PR #362, opened 2023-06-27
Update README.md

ask9027:
It is good if we use chroot as variable which more  convenient.


twaik:
`XKB_CONFIG_ROOT` can be deduced using `TMPDIR` so it can be stripped.

PR #470, opened 2023-09-24
Allow audio playback capture for Android 10+

kairusds:
This allows screen recorders on Android 10 and above to capture the audio

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -16,6 +16,7 @@
         android:supportsRtl="true"
         android:theme="@style/AppTheme"
         android:hardwareAccelerated="true"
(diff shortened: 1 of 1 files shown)


twaik:
Termux:X11 does not pplay any audio...

kairusds:
> Termux:X11 does not play any audio...

Iirc it plays audio when pulseaudio is configured

twaik:
Pulseaudio plays audio even when there is no termux-x11 process. It plays audio standalone with help of OpenSL ES API.

kairusds:
> Pulseaudio plays audio even when there is no termux-x11 process. It plays audio alone with help of OpenSL ES API.

Oh I didn't know that

PR #519, opened 2023-12-16
"Looper creation issue"

AnGelXoG:
I modified the code so that the image error no longer appeared.

Screenshot_20231127_080446

could you add the code? 

or you can simply solve the error on your own, I would greatly appreciate it <3


twaik:
The error does not appear in the case if you use latest versions of both loader and main application.

PR #543, opened 2024-01-27
Adapt to Samsung tab touchpad

mistletoe235:


twaik (changes requested):
Replacing constants used in code is not acceptable. It will break functionality for other Dex devices.

twaik:
I did not say I will not merge this. I requested changes.

mistletoe235:
I'm sorry that I'm not fimiliar with Andorid development, the lib that include the constants not include the 
public static final int CLASSIFICATION_TWO_FINGER_SWIPE = 3; 
However, another project I saw has it. And the file is read only, So I don't know how to change it.
image

twaik:
I am talking about something else. You changed `0x4000000` to other value without providing a fallback to other value. And the same thing with other code.

twaik:
> I'm sorry that I'm not fimiliar with Andorid development, the lib that include the constants not include the
> public static final int CLASSIFICATION_TWO_FINGER_SWIPE = 3

Probably not the best idea to inline constant, better to use the original name there.

>  can the 0x10000000 condition be added as just another else if block? so as to not change the existing 0x4000000 constant?

It can be used even in the same if block with logical OR.

Issue #593, opened 2024-04-04
[Bug]: Keyboard displayes in preference page


zongou:
### Problem description

Keyboard displays in preference page

### What steps will reproduce the bug?

1. Open app
2. Press Back key to show soft keyboard
3. Tap the preference button

### What is the expected behavior?

Do not show keyboard.

twaik:
I'll add some code to hide keyboard when becomes not focused.

zongou:
More details:
It not only happens in the within that app.
for example, if you are using google keyboard, after repeaing thoese steps but without tappping the preference button,
instead you go to google keyboard settings page, or android system settings page, it triggers that too.

zongou:
My idea to fix this bug:
Setting the flag to SHOW_IMPLICIT for general purpose, and then in the cases the soft keyboard cannot show, set it to SHOW_FORCED.

twaik:
Android framework does not report about IME status. It is shown or not. 
And no, we can not check ime size since
1. It is not possible before Android 10.
2. Because of IME appearing animation on some devices.

ask9027:
share screenshot , in my case there is no keyboard while open preferences as you told

zongou:
> share screenshot , in my case there is no keyboard while open preferences as you told


https://github.com/termux/termux-x11/assets/23202896/0f13f3ee-9c7d-4a24-990f-81f5a3a14f53

twaik:
@zongou check the latest build.

zongou:
> share screenshot , in my case there is no keyboard while open preferences as you told


https://github.com/termux/termux-x11/assets/23202896/032a56f8-60d8-42db-a212-709d42d7018d

zongou:
> @zongou check the latest build.

it indeed hides soft keyboard elsewhere, but i cannot switch keyboard now

https://github.com/termux/termux-x11/assets/23202896/99030c8b-8880-4481-86d2-0d505efb9009

twaik:
Try [this](https://github.com/termux/termux-x11/actions/runs/8598723904).

zongou:
> Try [this](https://github.com/termux/termux-x11/actions/runs/8598723904).

Currently this is no any issue i can find

PR #594, opened 2024-04-04
Fix: [Bug]: Keyboard displayes in preference page #593

zongou:


twaik:
What for?

twaik:
No. That will cause not-opening IME in some cases.

PR #601, opened 2024-04-11
chore: send mouse move via lorieMouseRelative device

KawaiiNahida:


twaik:
Did you see the `e.mouse.relative` condition? It checks if event was sent to relative device or not. It exists for a reason. And making code explicitly ignore this is not a option.

KawaiiNahida:
> Did you see the `e.mouse.relative` condition? It checks if event was sent to relative device or not. It exists for a reason. And making code explicitly ignore this is not a option.

I admit it's obviously not a good solution, but the current behavior of sending mouse scroll and movement via different devices causes the scroll to stop working when the mouse moves even slightly in most applications, it's so annoying.

I tried sending BUTTON_SCROLL to different devices, but I only received "RawMotion", and it seems that the applications do not respond to that. Additionally, I discovered that some devices' built-in touchpad gestures stop working when the pointer is captured by an app (for example, miui devices), and I will have to switch the pointer capture mode each time I switch between the mouse and built-in touchpad.

I've been running the modified version for almost a day and haven't found anything goes wrong with it, which is why I submitted the PR. I would appreciate it if there were other good solutions to solve this problem.

PR #617, opened 2024-04-28
Dex touchpad fix: orientation, click, gesture, etc.

### LINKED ISSUE #419: Touchpad right-click cancel and two-finger gesture axis inversion issue
Thank you for the new build.
Most of the Galaxy Tab S8+ keyboard cover touchpad compatibility is improved in the latest build. :)

Reports additional symptoms identified during the test.

https://github.com/termux/termux-x11/assets/101167173/c6a5627b-8a05-49c5-93d1-0b858b7da1ab

**1. touchpad right-click cancel**
If you click right on the touchpad, the click result seems to be canceled. (Video 00:09~00:20)

**2. two-finger gesture axis inversion issue**
eaacb2ff-e151-4059-ba1b-dcb517d4da2c

Two-finger gestures confirmed the symptoms of reversing the axis when moving up, down, left, and right. (video 00:22~00:30)

[log_20230808.txt](https://github.com/termux/termux-x11/files/12291566/log_20230808.txt)
The log file is here.

knyipab:
In short, it should fix #419. The suggested changes should have no impact on existing functionality. If you need to review the code changes and features added, I hope below explanation helps and won't be too long. 

## Root Causes of #419  
* Unlike touchscreen, the tocuchpad reports coordinates in its own (physical) orientation, does not flip along with display rotation.  
* Also, the comment at [line 185](https://github.com/termux/termux-x11/blob/5489f7c8726855cd55729660c2867139aa2a5a74/app/src/main/java/com/termux/x11/input/TouchInputHandler.java#L185) "Regular touchpads and Dex touchpad ... should be handled as touchscreens with **trackpad** mode" is not implemented as expected. Indeed, `mTouchpadHandler.handleTouchEvent()` will handle pointer inputs as hardware mouse, calling `mHMListener.onTouch` because of the "if condition" (`|| (event.getPointerCount() == 1 && mTouchpadHandler == null && (event.getSource() & InputDevice.SOURCE_TOUCHPAD) == InputDevice.SOURCE_TOUCHPAD)`) being positive. 

## My solution
* Remove the "if condition" so that touchpad event can invoke `TrackpadInputStrategy`. Note that we could not simply create another variable to be invoked because the code logics test `mInputStrategy` at multiple places. I found it easiest to just override `mInputStrategy` in DeX (`SamsungDexUtils.checkDeXEnabled(mContext)`). This makes sense  because (1) finger input in DeX can only be touchpad not touchscreen, and (2) mouse input handler in DeX is also overrided. 
* The flipped axis issue is tackled by introducing another option for captured pointer transformation, namely "Automatic (for touchpad)". Under this option, the rotation of display ~ID 1 (primary display I believe)~ will be checked and motion input will be handled accordingly. Hopefully this could handle both Pad and DeX touchpad issue. 
* The 3-finger down gesture to "toggle EK bar" is change to  "release pointer capture" under DeX mode. This is because (1) EK bar is useless and cannot be pressed in captured pointer mode, and (2) if user enters DeX using solely soft touchpad and soft keyboard, there should be a way to exit the capture mode and the more-than-2-finger gesture is just right play that role (3-finger as shortcut to release capture plus minimize window, 4-finger to only release capture). At the end, the current way for soft touchpad & soft keyboard under DeX to use EK bar is perhaps to first release pointer capture and use volume key to toggle EK bar.

### DIFF
--- a/app/src/main/java/com/termux/x11/input/TouchInputHandler.java
+++ b/app/src/main/java/com/termux/x11/input/TouchInputHandler.java
@@ -531,15 +531,8 @@ public class TouchInputHandler {
             if ((e2.getSource() & InputDevice.SOURCE_TOUCHPAD) == InputDevice.SOURCE_TOUCHPAD
                     && mInputStrategy instanceof InputStrategyInterface.TrackpadInputStrategy) {
                 float temp;
-                int transform = capturedPointerTransformation;
-                if (capturedPointerTransformation == CapturedPointerTransformation.AUTO) {
-                    for (Display display : mDisplayManager.getDisplays()) {
-                        if (display.getState() == Display.STATE_ON) {
-                            transform = display.getRotation() % 4;
-                            break;
-                        }
-                    }
-                }
+                int transform = capturedPointerTransformation == CapturedPointerTransformation.AUTO ?
+                        mDisplayManager.getDisplay(Display.DEFAULT_DISPLAY).getRotation() % 4 : capturedPointerTransformation;
                 switch (transform) {
                     case CapturedPointerTransformation.NONE:
                         break;
--- a/app/src/main/java/com/termux/x11/input/TouchInputHandler.java
+++ b/app/src/main/java/com/termux/x11/input/TouchInputHandler.java
@@ -247,6 +247,13 @@ public class TouchInputHandler {
     }
 
     public boolean handleTouchEvent(View view0, View view, MotionEvent event) {
+        // Regular touchpads and Dex touchpad (in captured mode) send events as finger too,
+        // but they should be handled as touchscreens with trackpad mode.
+        if (event.getToolType(event.getActionIndex()) == MotionEvent.TOOL_TYPE_FINGER &&
+                mTouchpadHandler != null && (event.getSource() & InputDevice.SOURCE_TOUCHPAD) == InputDevice.SOURCE_TOUCHPAD) {
+            return mTouchpadHandler.handleTouchEvent(view0, view, event);
+        }
+
         if (view0 != view) {
             int[] view0Location = new int[2];
             int[] viewLocation = new int[2];
@@ -279,11 +286,6 @@ public class TouchInputHandler {
             // Dex touchpad (in non-captured mode) sends events as finger, but it should be considered as a mouse.
             if (isDexEvent(event) && mDexListener.onTouch(view, event))
                 return true;
-
-            // Regular touchpads and Dex touchpad (in captured mode) send events as finger too,
-            // but they should be handled as touchscreens with trackpad mode.
-            if (mTouchpadHandler != null && (event.getSource() & InputDevice.SOURCE_TOUCHPAD) == InputDevice.SOURCE_TOUCHPAD)
-                return mTouchpadHandler.handleTouchEvent(view, view, event);
 
             // Give the underlying input strategy a chance to observe the current motion event before
             // passing it to the gesture detectors.  This allows the input strategy to react to the
@@ -368,11 +370,9 @@ public class TouchInputHandler {
     }
 
     public void setInputMode(@InputMode int inputMode) {
-        if (SamsungDexUtils.checkDeXEnabled(mActivity)) {
+        if (mTouchpadHandler == null)
             mInputStrategy = new InputStrategyInterface.TrackpadInputStrategy(mInjector);
-            return;
-        }
-        if (inputMode == InputMode.TOUCH)
+        else if (inputMode == InputMode.TOUCH)
             mInputStrategy = new InputStrategyInterface.NullInputStrategy();
         else if (inputMode == InputMode.SIMULATED_TOUCH)
             mInputStrategy = new InputStrategyInterface.SimulatedTouchInputStrategy(mRenderData, mInjector, mActivity);
--- a/app/src/main/java/com/termux/x11/input/TouchInputHandler.java
+++ b/app/src/main/java/com/termux/x11/input/TouchInputHandler.java
@@ -14,10 +14,12 @@ import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.graphics.PointF;
+import android.hardware.display.DisplayManager;
 import android.hardware.input.InputManager;
 import android.os.Handler;
 import android.os.Build;
 import android.util.DisplayMetrics;
+import android.view.Display;
 import android.view.GestureDetector;
 import android.view.InputDevice;
 import android.view.KeyEvent;
@@ -156,6 +158,7 @@ public class TouchInputHandler {
         mRenderData = renderData != null ? renderData :new RenderData();
         mInjector = injector;
         mActivity = activity;
+        mDisplayManager = (DisplayManager) mActivity.getSystemService(Context.DISPLAY_SERVICE);
 
         GestureListener listener = new GestureListener();
         mScroller = new GestureDetector(/*desktop*/ activity, listener, null, false);
@@ -365,7 +368,7 @@ public class TouchInputHandler {
     }
 
     public void setInputMode(@InputMode int inputMode) {
-        if (SamsungDexUtils.checkDeXEnabled(mContext)) {
+        if (SamsungDexUtils.checkDeXEnabled(mActivity)) {
             mInputStrategy = new InputStrategyInterface.TrackpadInputStrategy(mInjector);
             return;
         }
--- a/app/src/main/java/com/termux/x11/input/TouchInputHandler.java
+++ b/app/src/main/java/com/termux/x11/input/TouchInputHandler.java
@@ -301,18 +301,15 @@ public class TouchInputHandler {
             mTapDetector.onTouchEvent(event);
             mSwipePinchDetector.onTouchEvent(event);
 
+            // For hardware touchpad in DeX (captured mode), handle physical click buttons
+            if ((event.getSource() & InputDevice.SOURCE_TOUCHPAD) == InputDevice.SOURCE_TOUCHPAD) {
+                currentBS = event.getButtonState();
+                for (int[] button: buttons)
+                    if (isMouseButtonChanged(button[0]))
+                        mInjector.sendMouseEvent(null, button[1], mouseButtonDown(button[0]), true);
+                savedBS = currentBS;
+            }
             switch (event.getActionMasked()) {
-                case MotionEvent.ACTION_BUTTON_PRESS:
-                case MotionEvent.ACTION_BUTTON_RELEASE:
-                    // For hardware touchpad in DeX (captured mode), handle physical click buttons
-                    if ((event.getSource() & InputDevice.SOURCE_TOUCHPAD) == InputDevice.SOURCE_TOUCHPAD) {
-                        currentBS = event.getButtonState();
-                        for (int[] button: buttons)
-                            if (isMouseButtonChanged(button[0]))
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/termux/x11/MainActivity.java
```diff
@@ -176,7 +176,23 @@ protected void onCreate(Bundle savedInstanceState) {
         mInputHandler = new TouchInputHandler(this, new RenderStub.NullStub() {
             @Override
             public void swipeDown() {
-                toggleExtraKeys();
+                if (SamsungDexUtils.checkDeXEnabled(getApplicationContext())) {
+                    lorieView.releasePointerCapture();
+                    lorieParent.releasePointerCapture();
```
twaik, line 181: No need to ask for pointer release twice. Pointer capture is activity-wide.

app/src/main/java/com/termux/x11/MainActivity.java
```diff
@@ -176,7 +176,23 @@ protected void onCreate(Bundle savedInstanceState) {
...
+                }
+            }
+            @Override
+            public void swipeUp() {
+                // Gesture for soft touchpad in DeX mode that cannot trigger KEYCODE_BACK easily
+                if (SamsungDexUtils.checkDeXEnabled(getApplicationContext())) {
+                    toggleKeyboardVisibility(MainActivity.this);
+                    Log.d("MainActivity", "Toggling keyboard visibility");
+                    if(inputMethodManager != null) {
+                        inputMethodManager.showSoftInput(lorieView, InputMethodManager.SHOW_FORCED);
```
twaik, line 193: Is there a reason to trigger toggling IME twice?

app/src/main/java/com/termux/x11/input/TouchInputHandler.java
```diff
@@ -361,6 +397,35 @@ private class GestureListener extends GestureDetector.SimpleOnGestureListener
...
+                float temp;
+                int transform = capturedPointerTransformation;
+                if (capturedPointerTransformation == CapturedPointerTransformation.AUTO) {
+                    for (Display display : mDisplayManager.getDisplays()) {
+                        if (display.getState() == Display.STATE_ON) {
+                            transform = display.getRotation() % 4;
+                            break;
+                        }
+                    }
+                }
```
twaik, line 413: Is there any specific reason to obtain display via `DisplayManager` and not using `mActivity.getLorieView().getDisplay()`? This definitely will break in multi-display mode.

knyipab: Cuz in smarphone HDMI external display DeX mode (using phone as software touchpad), the transformation should depend on phone orientation (=soft touchpad) but `mActivity.getLorieView().getDisplay()` return HDMI display which is useless. 
This enumeration works well for case of both external display DeX + software touchpad and , tablet hardware touchpad. 

I do reckon the risk of not working well in some more complex setup, such as external display DeX + hardware touchpad, or Tablet in external display DeX + hardware touchpad, and perhaps more etc. 

A better approach might be depends on input device information. But I don't have both devices so I used that workaround. FYR. soft touchpad might be "sec_touchpad" but I am not so sure. In case you want to implement: 

Screenshot from 2024-05-17 08-01-47

app/src/main/java/com/termux/x11/MainActivity.java
knyipab: I just copied from somewhere else without much thinking about it. Feel free to amend if you think appropriate

app/src/main/java/com/termux/x11/MainActivity.java
knyipab: During my previous test, `toggleKeyboardVisibility` did not work. However, I cannot reproduce it now, so I think one `toggleKeyboardVisibility` should just work. 

I was mistaken in previous test perhaps because the keyboard toggling feature disappear sometimes when the app switches from phone screen to external display or vice versa. It isn't too annoying so now I simply close and re-open everytime.

twaik:
> Remove the "if condition" so that touchpad event can invoke TrackpadInputStrategy. Note that we could not simply create another variable to be invoked because the code logics test mInputStrategy at multiple places. I found it easiest to just override mInputStrategy in DeX (SamsungDexUtils.checkDeXEnabled(mContext)). This makes sense because (1) finger input in DeX can only be touchpad not touchscreen, and (2) mouse input handler in DeX is also overrided.

The touchpad handling code is created for both Dex and real touchpad cases. And there may be physical touchpad, mice and touchpad connected to the same device at the same time. That is a reason I am invoking the code handling touchpads and Dex in separate TouchInputHandler instance with forced trackpad mode. mInputStrategy overriding will break this case.

knyipab:
I am indeed well aware of the fact that touchpad + mouse should be fine, but in touchpad + mouse + touchscreen (in DeX) or perhaps simply touchscreen in DeX, the touchscreen mode would be completely overridden. Besides, I am not sure how touchscreen can behave under onscreen DeX mode (as a touchpad? or mouse? does Termux:X11 touchscreen mode work?) which is only available to Tablet user and I don't own one. 

It is just that 

> the code logics test `mInputStrategy` at multiple places

So it will require a more changes to your code base (probably a real `mTouchpadHandler` resembling touchpad InputStrategy). I was afraid to do such big change and unsure about my expertise. 

It's up to you. Want me to implement it (though not any time soon) or you change it directly? Or is it still acceptable as is?

twaik:
Ok, probably it is a good time to make a rebase.

> So it will require some more changes to your code base

In the case if you detect event came from touchpad you should redirect it to mTouchpadHandler immediately (in TouchInputHandler::onTouch). 

About
```
                if (capturedPointerTransformation == CapturedPointerTransformation.AUTO) {
                    for (Display display : mDisplayManager.getDisplays()) {
                        if (display.getState() == Display.STATE_ON) {
                            transform = display.getRotation() % 4;
                            break;
                        }
                    }
                }
```
I still think it is a good idea to use `DisplayManager::getDisplay(Display.DEFAULT_DISPLAY)` which should return phone's primary display.

About
```
                case MotionEvent.ACTION_BUTTON_PRESS:
                case MotionEvent.ACTION_BUTTON_RELEASE:
                    // For hardware touchpad in DeX (captured mode), handle physical click buttons
                    if ((event.getSource() & InputDevice.SOURCE_TOUCHPAD) == InputDevice.SOURCE_TOUCHPAD) {
                        currentBS = event.getButtonState();
                        for (int[] button: buttons)
                            if (isMouseButtonChanged(button[0]))
                                mInjector.sendMouseEvent(null, button[1], mouseButtonDown(button[0]), true);
                        savedBS = currentBS;
                        break;
                    }
```
I still think it is a good idea to move it out from switch. There are some weird firmwares with custom actions hardcoded. Sending event will not be triggered in the case if button state is not changed so it should be safe.

Also be aware of new preferences making possible to configure three-finger swipes, back button, volume keys actions. Also there is new option to toggle soft keyboard from additional keyboard bar (enabled by default) so the changes overriding default actions in dex mode should be removed.

knyipab:
Rebased and addressed all your comments. 

> In the case if you detect event came from touchpad you should redirect it to mTouchpadHandler immediately (in TouchInputHandler::onTouch).

Sounds smart and requires minimal changes. I frontload the `mTouchpadHandler` in `handleTouchEvent` in https://github.com/termux/termux-x11/pull/617/commits/a5a91473a0224f9c8391f970147ddd917affdae1.  Now `mInputStrategy` overriding only takes place when it is `mTouchpadHandler`. Perhaps a thorough test is probably having a tablet owner to use physiccal touchpad and touchscreen in non-trackpad mode together. 

> I still think it is a good idea to use `DisplayManager::getDisplay(Display.DEFAULT_DISPLAY)` which should return phone's primary display.

Thanks. While `mActivity.getLorieView().getDisplay()` does not work well for my foldable device soft touchpad under DeX, `DisplayManager::getDisplay(Display.DEFAULT_DISPLAY)` works perfectly. 

> I still think it is a good idea to move it out from switch. There are some weird firmwares with custom actions hardcoded. Sending event will not be triggered in the case if button state is not changed so it should be safe.

I moved it out of the switch statement. See if you may want to invite other users to conduct a final UAT cuz I don't own a tablet. 

> Also be aware of new preferences making possible to configure three-finger swipes, back button, volume keys actions. Also there is new option to toggle soft keyboard from additional keyboard bar (enabled by default) so the changes overriding default actions in dex mode should be removed.

I am fine with new preferences and now this PR leaves the three-finger swipe gestures untouched. Cuz I found that under DeX soft touchpad mode, I can escape the pointer capture by swipe up at the bottom of phone screen. That will bring up the bottom "navigation gesture bar" (not sure the correct name) and escape pointer capture. Nonetheless, it assumes new Termux:X11 user to know such trick, and perhaps  some first time user would be panic and ends up unplugging the HDMI cable to escape. 

Well... perhaps this app has already made quite some assumptions about new users when we don't have a starter guide LOL. For example, I discovered preferences page can be triggered from notification bar after few weeks of use, or perhaps I am stupid haha.

knyipab:
By the way, 
1. shall we make "transform captured pointer movements" by default to be "Automatic (for touchpad)"? It seems harmless and perhaps enhane user experience
2. would it be better to show the current chosen option right below "transform captured pointer movements"? I couldn't quite find where to implement unless dive deeper in your code.

twaik:
> Well... perhaps this app has already made quite some assumptions about new users when we don't have a starter guide LOL. For example, I discovered preferences page can be triggered from notification bar after few weeks of use, or perhaps I am stupid haha.

Not stupid. I have some troubles with writing user manuals. 
Also yesterday I pushed a commit which adds `preferences` and `toggle soft keyboard` actions to the additional key bar.

> shall we make "transform captured pointer movements" by default to be "Automatic (for touchpad)"? It seems harmless and perhaps enhance user experience

I already checked it when tried to fix touchpad by myself. It breaks touchpads on other devices.

> would it be better to show the current chosen option right below "transform captured pointer movements"? I couldn't quite find where to implement unless dive deeper in your code.

I think it will be enough to change the summary of preference when it changes. I'll do that a bit later.

twaik:
Probably polling for display rotation with `mDisplayManager.getDisplay(Display.DEFAULT_DISPLAY).getRotation()` on every single event is not very wise. You can get more than 200 input events in a sec and invoking IPC for every single event seems to be wasteful. Probably it would be better to create self-calling lambda like `checkXEvents` which invokes `MainActivity.handler.postDelayed` to call itself every 300 ms which will save current rotations somewhere.

twaik:
@AlphaBs @hansm629 can you please test latest CI artifact of this PR?

hansm629:
> @AlphaBs @hansm629 can you please test latest CI artifact of this PR?

@twaik
I'll test it out after work today

AlphaBs:
I tested [this build](https://github.com/termux/termux-x11/actions/runs/9737702734) on Galaxy Tab S8+ bookcover keyboard, with trackpad, capturing external pointer, and automatic point movements mode.

Touch gestures (tap, two-finger tap to right click, two-finger scroll, click buttons, long tap to drag, tap-to-move), scroll axis, pointer movement axis, pointer speed are all normal and work perfectly. Thanks for a great job!

knyipab:
@twaik just pushed a commit to update rotation using  `DisplayManager#registerDisplayListener()` which works on my phone. See if you may have concern about firmware/vendor-specific issue and prefer `MainActivity.handler.postDelayed`.

Issue #605, opened 2024-04-20
[feature request] option to factor in the height of extra keys bar into display dimension calculation


knyipab:
The extra keys bar is currently shown on top of the display. The bar could indeed block what can be seen at the bottom of display, such as when browsing websites in a maximized browser window. It could cause inconvenience for users who want to have the bar always on, or he/she would need to hide the bar every time to see the bottom of display and then toggle the bar again. 

Therefore, it would be really great to have an option to make the app to recalculate display dimension so not to overlap with the extra keys bar whenever the bar is toggled. Thank you.

twaik:
The extra key bar was designed to be toggled on or off as needed. Intentionally. For convenient switching, you can use two gestures: a three-finger swipe or the "volume down" button (activated in the settings).

Recalculating size of EK bar and screen surface was used in older versions but it was removed because it caused flickering of screen and there is no fix for this. It will not be back again.

RalfWerner:
@knyipab The same function is also possible in the Termux app. There the **EK** with `LogTap` is toggled in the drawer - mostly. The **EK** is displayed there above the On-Screen KB. The Termux prompt is displayed above the **EK**. So is never covered.
In version `1.02.04` of Termux-x11, the same behavior has been implemented and some other functions have also been lost in `1.03.00`. However, there are also good extensions in the latest artifact. For me, the advantages of `1.02.04` predominate on _pure phones_ without additional input devices.

knyipab:
Similar to #607. Do you mind me putting this feature in a sub-page called "unstable features"?

Still, this option can be useful. Even with three-finger gesture, using the arrow buttons the navigate the terminal at the bottom of VS Code is a pain. I am happy to make a pull request make available an option under "unstable features".

RalfWerner:
I would like that very well, how I wrote in my deleted commit above.

knyipab:
Inspired by RealVNC viewer, an option for opacity of EK bar through `.setAlpha()` can be a good alternative to PR #618 , at least solving the pain point of using vs code terminal. I could implement, subject to your view.

twaik:
Probably it is a nice idea.

twaik:
Closed by #618

PR #618, opened 2024-04-30
option to adjust display height for EK bar

knyipab:
A minimalist implemention of #605

Tested on my samsung fold 5, and no flickering seen. 

Depending on your preference, if needed, it could be added in description "may cause flickering for some devices" to prevent user from opening issues on github.

### DIFF
--- a/app/src/main/java/com/termux/x11/MainActivity.java
+++ b/app/src/main/java/com/termux/x11/MainActivity.java
@@ -631,6 +631,7 @@ public class MainActivity extends AppCompatActivity implements View.OnApplyWindo
                 parent.removeView(pager);
                 parent.addView(pager, 0);
             }
+            frm.setPadding(0, 0, 0, preferences.getBoolean("adjustHeightForEK", false) && show ? pager.getHeight() : 0);
 
             if (enabled && saveState) {
                 SharedPreferences.Editor edit = preferences.edit();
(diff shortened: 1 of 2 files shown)


RalfWerner:
@knyipab You did a fantastic job! With this version, the behavior of version **1.02.04** is partially restored. However, the `opacityEKBar` option is missing here. In some cases it would useful.
image

The implementation of the keys into [symbols](https://github.com/termux/termux-x11/assets/45426380/22093757-657a-4649-bcd8-69c31156551c) has been incomplete (**red**) since version **1.03.00**. Here a synchronized script for the EK is used in both apps. If you could correct that, it would be almost perfect for the EK. I would prefer the exclusive effect of `hideEKOnVolDown` = **true**.

I edit [Xclients](https://github.com/termux/termux-x11/assets/45426380/333595c3-d890-49ef-997f-b8e4f7686918) who use Direct mode as `touchMode`, where the use of several fingers is the typical case. With every movement of more than two fingers downwards (in X11-App), the EK is toggled on/off. If that would not be done at **true**, this problem would also be solved.

RalfWerner:
@knyipab I wanted to thank you and contribute a few ideas. Unfortunately, the commit (above) was deleted by `termux`. Do you have a channel where this is not done?

knyipab:
I just opened the issues function of [my fork](https://github.com/knyipab/termux-x11) and you may try.

twaik:
> Recalculating size of EK bar and screen surface was used in older versions but it was removed because it caused flickering of screen and there is no fix for this.

https://github.com/termux/termux-x11/assets/9674930/acd756b4-119c-4dcd-9c89-e637ebb4085f

Issue #147, opened 2022-08-09
Lack of Chinese input method, consider to add fcitx5


EdwardZhangLink:
I have installed `TeXLive` and `texstudio`, but I lack a proper Chinese input method to write the documentation. I need a Chinese input method like `fcitx5` to add to the library.

Fcitx5 Wiki Page
[https://fcitx-im.org/wiki/Fcitx_5](https://fcitx-im.org/wiki/Fcitx_5) 
[https://fcitx-im.org/wiki/Fcitx_5/zh-cn](https://fcitx-im.org/wiki/Fcitx_5/zh-cn) 
Fcitx5 Source Code Page
[https://github.com/fcitx/fcitx5](https://github.com/fcitx/fcitx5)

twaik:
I know what is fcitx5, but it will not help. 
Android sends proper keycodes only in the case if you input english characters, in any other case it sends Unicode characters. 
Xwayland can receive input only as linux scancodes. 
I am planning to make additional X client based on `x11vnc` and `xdotool` code which will be able to input every single Unicode symbol.

In the case if you can not wait you can try to use any **linux (not Android)** on screen keyboard solution which supports chinese.

EdwardZhangLink:
> I know what is fcitx5, but it will not help. Android sends proper keycodes only in the case if you input english characters, in any other case it sends Unicode characters. Xwayland can receive input only as linux scancodes. I am planning to make additional X client based on `x11vnc` and `xdotool` code which will be able to input every single Unicode symbol.
> 
> In the case if you can not wait you can try to use any **linux (not Android)** on screen keyboard solution which supports chinese.

First of all, thank you for your reply. I am indeed a bit frustrated. I tried to understand GUI related knowledge, but now they are too complicated for me. Maybe I didn't understand your reply, maybe I didn't express it clearly, actually I don't intend to send Chinese characters to `xwayland` directly through Android keyboard, but give Chinese candidate words in pinyin form by inputting English characters through input method. I think `fcitx5` will work like this, but maybe not so simple, hope I can get some help.
Besides, I found it support `wayland` in `fcitx5` related page, but I don't know if it is helpful. Because the special feature of Chinese input is always to use an input method to finalize Chinese input by giving Chinese character candidates, but Linux on-screen keyboard obviously doesn't have such function. Maybe we can consider to adapt package like `fcitx5` first, and then test whether it is applicable.
Finally, looking forward to your progress!

junrenshi:
[...]

Issue #115, opened 2022-04-17
Can't input Chinese


dontknowhy:
It seems like termux-x11 Nnot supported Chinese
but i can input Chinese in VNC Server
why?

If Xwayland does not support X input method for input Chinese, please close this issue, if so, please let termux-x11 also support
input Chinese

twaik:
Duplicate #147

twaik:
Check latest build (which is "[Input of custom unicode characters.](https://github.com/termux/termux-x11/actions/runs/4333077282)".

PR #620, opened 2024-05-02
Enable Gboard CJK keyboards

knyipab:
~Relates to #115 and #147~. 

Add a preference option "Workaround to enable CJK Gboard". 

* The issue of Gboard IME showing up as QWERTY affects mostly only CJK keyboards.
* Tested on my Samsung Z Fold5, Android 14. Gboard 14.1.04.621126403-release-arm64-v8a. 
* It is made as a preferences.xml option because the impact on other IMEs is not well known and there may be side effects. API to detect current IME in use `InputMethodManager#getCurrentInputMethodInfo()` is not available until recent API level (34).

### DIFF
--- a/app/src/main/java/com/termux/x11/LorieView.java
+++ b/app/src/main/java/com/termux/x11/LorieView.java
@@ -31,8 +31,6 @@ import com.termux.x11.input.InputStub;
 import com.termux.x11.input.TouchInputHandler;
 
 import java.nio.charset.StandardCharsets;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.TimeUnit;
 import java.util.regex.PatternSyntaxException;
 
 @Keep @SuppressLint("WrongConstant")
@@ -53,7 +51,7 @@ public class LorieView extends SurfaceView implements InputStub {
     private InputMethodManager mIMM = (InputMethodManager)getContext().getSystemService( Context.INPUT_METHOD_SERVICE);
     private String mImeLang;
     private boolean mImeCJK;
-    private boolean enableGboardCJK;
+    public boolean enableGboardCJK;
     private Callback mCallback;
     private final Point p = new Point();
     private final SurfaceHolder.Callback mSurfaceCallback = new SurfaceHolder.Callback() {
@@ -295,13 +293,7 @@ public class LorieView extends SurfaceView implements InputStub {
         if (mIMM.getCurrentInputMethodSubtype().getLanguageTag().length() >= 2 && !mIMM.getCurrentInputMethodSubtype().getLanguageTag().substring(0, 2).equals(mImeLang))
             mIMM.restartInput(this);
         else if (recheck) { // recheck needed because sometimes requestCursorUpdates() is called too fast, before InputMethodManager detect change in IM subtype
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S)
-                CompletableFuture.delayedExecutor(40, TimeUnit.MILLISECONDS).execute(() -> { checkRestartInput(false); });
-            else
-                new Thread(() -> { try {
-                    Thread.sleep(40);
-                    checkRestartInput(false);
-                } catch (Exception e) { System.err.println(e); } }).start();
+            MainActivity.handler.postDelayed(() -> checkRestartInput(false), 40);
         }
     }
     @Override
--- a/app/src/main/java/com/termux/x11/LorieView.java
+++ b/app/src/main/java/com/termux/x11/LorieView.java
@@ -11,6 +11,7 @@ import android.graphics.Color;
 import android.graphics.Point;
 import android.graphics.Rect;
 import android.graphics.drawable.ColorDrawable;
+import android.os.Build;
 import android.text.InputType;
 import android.util.AttributeSet;
 import android.util.Log;
@@ -18,8 +19,10 @@ import android.view.KeyEvent;
 import android.view.Surface;
 import android.view.SurfaceHolder;
 import android.view.SurfaceView;
+import android.view.inputmethod.BaseInputConnection;
 import android.view.inputmethod.EditorInfo;
 import android.view.inputmethod.InputConnection;
+import android.view.inputmethod.InputMethodManager;
 
 import androidx.annotation.Keep;
 import androidx.annotation.NonNull;
@@ -28,6 +31,8 @@ import com.termux.x11.input.InputStub;
 import com.termux.x11.input.TouchInputHandler;
 
 import java.nio.charset.StandardCharsets;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
 import java.util.regex.PatternSyntaxException;
 
 @Keep @SuppressLint("WrongConstant")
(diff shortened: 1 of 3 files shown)


knyipab:
Incorrect text UTF8 text sometimes sent to X clients. Not a bug of this commit, instead there are likely some problems in `mInjector.sendTextEvent()`. Lemme investigate and will reopen after fixing.

knyipab:
Cannot reproduce the aforementioned bug after re-run`termux-x11` command in Termux. In addition, it's not a bug of this PR, so I re-open this PR. 

**Only for record**: as far as I can tell, there may be bug in XCB receiver side. Log by `Java_com_termux_x11_LorieView_sendTextEvent` shows that correct character is sent. However, incorrect characters shown in XFCE DE is some characters that was typed before. I tested that `termux-x11` session for more than a day and perhaps changes some internal state of XCB receiver side, causing the bug. Anyways, I can't reproduce that at the moment.

knyipab:
Sorry for confusion. I will stop editing here and look for @twaik your comments. 

More testing are done as below. 

| Device | Android Ver. | GBoard Ver. | LangTag detection to activate CJK Gboard |
| --- | --- | --- | --- |
| Samsung Z Fold5 | 14 | 14.1.04.621126403-release-arm64-v8a | Working |
| Samsung S21  | 14 | 14.0.11.612796517-release-arm64-v8a | Working |
| Pixel 4a | 13 | 14.0.11.612796517-release-arm64-v8a | Not working, blank `getLanguageTag()` |
| Pixel 4a | 13 | 14.1.04.621126403-release-arm64-v8a | Not working, blank `getLanguageTag()` |
| Emulated (AVD Pixel 4) | 13 | 14.1.04.621126403-release-x86_64 | Not working, blank `getLanguageTag()` |
| Emulated (AVD Pixel 8) | 14 | 12.4.05.482060964-preload-x86_64 | Not working, blank `getLanguageTag()` and `requestCursorUpdates()` not even called |
| Emulated (AVD Pixel 8) | 14 | 14.1.04.621126403-release-x86_64 | Working |

This workaround is by no mean elegant but probably the only existing workaround siince being last discussed in 2021 https://github.com/termux/termux-app/issues/1539. Bit more context of the problem and welcome smarter workaround: the `TYPE_TEXT_VARIATION_NORMAL` can enable CJK, emoji and clipboard input from Gboard, but at the same time activates Gboard word suggestion causing backspace not working and the word suggestion can be suppressed through `restartInput()`/`invalidateInput()`, but does not work well for non-CJK Gboard such as English (when type fast and two keypresses `commitText()` so close in time, one of them will be ignored). Therefore, the `getLanguageTag()` detection is a workaround to set `TYPE_TEXT_VARIATION_NORMAL` only when it is needed (i.e. CJK IME). 

Grateful for merging given importance to Gboard CJK users. I don't mind putting it onto an "Experimental" section of perefernces.xml. Thanks.

twaik:
> Relates to #115 and #147.

Not related. 
That happened because Xwayland (used a long time ago) did not have any instruments to input custom Unicode symbols which are not in current keymap.

> Only for record: as far as I can tell, there may be bug in XCB receiver side. 

There is no "XCB receiver" in termux-x11. There is some custom light weight protocol to avoid XCB dependency (which adds 1MB+ for every architecture variant and 4.5MB to universal build).

> However, incorrect characters shown in XFCE DE is some characters that was typed before.

Termux:X11 sends characters the same way tigervnc-server and x11vnc do. It maps characters to keycodes with no associated keysyms (https://github.com/termux/termux-x11/blob/6fbd33211c8c60be852bb9f0023f735774399611/app/src/main/cpp/lorie/InputXKB.c#L52-L54) and sends key event to application. This method is not really reliable but it works for most text editors and does not interfere with existing hotkey handling implementations.

I did not find better way to implement this, x11vnc way is the best option I found.

knyipab:
Thanks. Since the bug is really not a problem of this PR and neither Gboard, I opened #625  for test result and further discussion. 

Back to this PR, I tested together with #617 (which makes available the 3-finger swipe up gesture to activate soft keyboard in DeX mode) and commit another workaround for DeX mode in which `InputConnection#requestCursorUpdates()` is not called by Gboard at all. 

Any adjustment to this PR you would like me to do before merging?

813ethan:
any chance this would be merged? working perfectly for me with xfce4 for this branch with traditional chinese gboard

twaik:
Currently I am working on some other changes, I'll review it again a bit later. Also both this and #617 must be rebased (and the latter one has unresolved reviews).
Termux:X11 code has been pretty much changed since prs were opened.

knyipab:
Rebased. 

The `Alt` issue can be resolved by user going to Gboard settings to disable "Emoji with physical keyboard / Physical Alt to show emoji keyboard".

knyipab:
Indeed, for those Gboard users who turn off suggestion and autocompletion for their English keyboard (and vaguely other letter-based IMEs), the lang subtype detection (`mIMM.getCurrentInputMethodSubtype().getLanguageTag()`) and `mImeCJK` would be unnecessary and `LorieView` can simply set `InputType.TYPE_TEXT_VARIATION_NORMAL`, and then users can enjoy voice input and emoji input. However, I am not that kind of no-suggestion & no-autocompletion user. I am just saying it and see if you may want to extend an option for those users to force `InputType.TYPE_TEXT_VARIATION_NORMAL`. 

Thanks for your time and consideration.

twaik:
Pay attention to the review please.

twaik:
Probably it can be fixed bytusing IBus methods. AFAIK it can let us send unicode symbols without mapping them to keys and avoid this weird behaviour.

knyipab:
It seems to require ibus to be on the deps of `termux-x11-nightly` package and start an ibus daemon. Is that right?

Another info: previous test results in #625 show that the implementation `tigervnc` + realvnc viewer can be 100% reproducible and output is reliable, unlike Termux:X11's unpredictable weird output. (Grateful for letting me know if their approach has sacrificed any features, but) Their approach is almost perfect except that the support to some specific language characters is lacking, which doesn't seem to me to be an implementation limitation. Anything we can possibly learn from their approach?

twaik:
It is pretty much weird, because I compared termux-x11's implementation with the original one and did not find any differences except function names, dropped upper-lower feature in lorieAddKeysym (because most likely there will be no difference), pressKey replaced by QueueKeyboardEvents, some logging stuff changed, dropped Shift+Alt mapping to META stuff. But generally they are equal.

knyipab:
No expertise, perhaps about the frontend side (RealVNC) (??). 

But I think clues can be found in the `xev` log files in #625: [x11-zh-sc.log](https://github.com/termux/termux-x11/files/15262667/termux-x11-zh-sc.log) and [vnc-zh-sc.log](https://github.com/termux/termux-x11/files/15262694/tigervnc-realvnc-zh-sc.log). One observable difference is that termux-x11 sends some events between each characters: 
```
MappingNotify event, serial 51, synthetic NO, window 0x0,
    request MappingKeyboard, first_keycode 203, count 1

MappingNotify event, serial 51, synthetic NO, window 0x0,
    request MappingKeyboard, first_keycode 203, count 1
```
There may be more difference but I haven't probed into.

twaik:
MappingEvent is event telling programs to associate some keycode (one in this case) to specific unicode symbol.

knyipab:
Nonetheless, seems that no `MappingNotify` (but `MapNotify`) is seen in the log of `tigervnc`. Where is `MappingNotify` produced in the termux-x11 code base? I can remove it and test.

twaik:
Main entry point for unicode input is `lorieKeysymKeyboardEvent`. But in the case tou remove it keyboard input will be possible only with `Prefer scancodes` option.

PR #627, opened 2024-05-10
EK bar opacity

knyipab:
Implement EK bar opacity, for #605 .


RalfWerner:
@knyipab, can I get an artifact from you, if you've success?
<details><summary> various checks with extra keys </summary>

image
</details>

knyipab:
@RalfWerner true that (even with EK bar opacity), non-trackpad still have to deactivate EK bar to access the bottom area. Still, it provides visibility. 
@twaik I don't hold strong view between this PR #627 opacity and PR #618 adjust height. It's good to at least have a choice for user in the preferences page.

PR #650, opened 2024-06-05
Fix touchpad transform having no effect

TudbuT:
The Touchpad transform currently has no effect because the TouchInputHandler mTouchpadHandler does not receive updates to the config. I looked through the uses of it and found out that although one could apply a far simpler patch, the mTouchpadHandler actually has no extra functionality and can be removed without any change in behavior. This both reduces code complexity and fixes the touchpad transform. 

If this solution is unwanted, I can move to the smaller patch instead - though I personally don't see any metric where that solution would be better unless mTouchpadHandler has some very obscure function that I wasn't able to find or discover during preliminary testing.


TudbuT:
Alright, that makes sense I suppose. Hold on.

TudbuT:
That should fix it now.

twaik:
#650 so I am closing this. Termux's Gimp problem is not related to termux-x11 plugin so I am closing this.

PR #667, opened 2024-07-04
Escape and Ctrl+Alt keys in user action section

knyipab:
See if this looks good to you. Feel free to change. Thanks. 

Btw, I thought of changing the name of "Pause key intercepting with Esc key" option cuz that's not so accurate. Indeed, any gestures/keys/buttons that release pointer will be affected by this option. Probably it's better to remove this option, and then add "release pointer" to `userActionsValues`, so to distinguish from "release pointer and keyboard capture". That involes more changes to the code base, so I didn't implement and leave it to your decision.

### DIFF
--- a/app/src/main/java/com/termux/x11/input/InputEventSender.java
+++ b/app/src/main/java/com/termux/x11/input/InputEventSender.java
@@ -18,6 +18,7 @@ import com.termux.x11.MainActivity;
 
 import java.util.List;
 import java.util.TreeSet;
+import java.util.function.Consumer;
 
 /**
  * A set of functions to send users' activities, which are represented by Android classes, to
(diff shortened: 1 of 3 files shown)


knyipab:
Screenshot_20240705_005353_TermuxX11

knyipab:
Rebased
Screenshot_20240705_230333_TermuxX11

twaik:
I do not think it is a good idea to do things like this for key combinations. Probably it would be better to choose something like Host key (like in Virtualbox or VMWare Workstation) (and probably make it configurable, defaulting to right control) and implement additional key combinations for different actions (like toggling EK bar, fullscreen, other stuff.

knyipab:
Umm... I used only virt-manager and after looking at some screenshots, vmware workstation's model (any/combo of Ctrl, Shift, Alt and Meta as "host key") is probably easier to implement cuz Android natively provides `isCtrlPressed()`, `isShiftPressed()`, `isAltPressed()` and `isMetaPressed()`, so the current PR code logic won't need to change substantially. User can add entries of Host key + any additional key (though need to figure out how to do such variable-sized config layout & storage). Lastly, "Escape key" will remain to be a separate item in this preference page. What do you think?

robertkirkman:
Yes I enabled "Capture external pointer devices" and now I am stuck in Termux:X11 and I can't escape because I forgot to install a terminal emulator, so I can't use `pkill termux-x11`

image

robertkirkman:
I clicked on "log out" in MATE but the cursor is still captured by Termux:X11 so now I really can't do anything

image

robertkirkman:
This is using master branch not this PR but the reason why I posted it here is because it sounds like this PR might give a way to fix this but I don't know

Issue #665, opened 2024-07-02
[Bug]: ctrl-space-workaround does not work


natebragg:
### Problem description

Some systems don't handle CTRL+space correctly.  On my machine, running `xev` from within `termux-x11 -xstartup "aterm"`, pressing and releasing left CTRL results in the following output:

```
KeyPress event, serial 32, synthetic NO, window 0x600001,
    root 0x511, subw 0x0, time 10088320, (1466,1198), root:(1467,1199),
    state 0x0, keycode 37 (keysym 0xffe3, Control_L), same_screen YES,
    XLookupString gives 0 bytes:
    XmbLookupString gives 0 bytes:
    XFilterEvent returns: False

KeyRelease event, serial 32, synthetic NO, window 0x600001,
    root 0x511, subw 0x0, time 10089266, (1466,1198), root:(1467,1199),
    state 0x4, keycode 37 (keysym 0xffe3, Control_L), same_screen YES,
    XLookupString gives 0 bytes:
    XFilterEvent returns: False
```

Pressing left CTRL followed by pressing the space bar results in the following output:

```
KeyPress event, serial 32, synthetic NO, window 0x600001,
    root 0x511, subw 0x0, time 10091627, (1466,1198), root:(1467,1199),
    state 0x0, keycode 37 (keysym 0xffe3, Control_L), same_screen YES,
    XLookupString gives 0 bytes:
    XmbLookupString gives 0 bytes:
    XFilterEvent returns: False

KeyRelease event, serial 32, synthetic NO, window 0x600001,
    root 0x511, subw 0x0, time 10092033, (1466,1198), root:(1467,1199),
    state 0x4, keycode 37 (keysym 0xffe3, Control_L), same_screen YES,
    XLookupString gives 0 bytes:
    XFilterEvent returns: False
```

These are the same except for the time field.  This same issue is present in normal termux, and is fixed using the `ctrl-space-workaround` setting inside the `.termux/termux.properties` file.

Please include a setting that fixes the same issue.

### What steps will reproduce the bug?

Follow the steps above.

### What is the expected behavior?

CTRL+space is processed properly, and gives the correct keycodes inside `xev`.

twaik:
`.termux/termux.properties` is a properties file of termux-app, not related to termux-x11.
Try to disable `Hardware keyboard scancodes workaround` in preferences.

natebragg:
[...]

PR #671, opened 2024-07-08
OpenXR - BT/OTG support and hardcode native resolution

lvonasek:
This addresses XR issues mentioned in https://github.com/termux/termux-x11/issues/665.

* Support XR mode without controllers (patch of `AndroidManifest.xml`)
* Set resolution for "native" lower (patch of `LorieView.java`)
* Support for BT/OTG keyboard (patch of `XrActivity.java`)

### DIFF
--- a/app/src/main/java/com/termux/x11/LorieView.java
+++ b/app/src/main/java/com/termux/x11/LorieView.java
@@ -140,7 +140,7 @@ public class LorieView extends SurfaceView implements InputStub {
         int height = getMeasuredHeight();
         if (XrActivity.isEnabled()) {
             width = 1024;
-            height = 1024;
+            height = 768;
         }
         int w = width;
         int h = height;
--- a/app/src/main/java/com/termux/x11/LorieView.java
+++ b/app/src/main/java/com/termux/x11/LorieView.java
@@ -138,6 +138,10 @@ public class LorieView extends SurfaceView implements InputStub {
         Prefs prefs = MainActivity.getPrefs();
         int width = getMeasuredWidth();
         int height = getMeasuredHeight();
+        if (XrActivity.isEnabled()) {
+            width = 1024;
+            height = 1024;
+        }
         int w = width;
         int h = height;
         switch(prefs.displayResolutionMode.get()) {
@@ -180,8 +184,11 @@ public class LorieView extends SurfaceView implements InputStub {
         if (prefs.displayStretch.get()
               || "native".equals(prefs.displayResolutionMode.get())
               || "scaled".equals(prefs.displayResolutionMode.get())) {
-            getHolder().setSizeFromLayout();
-            return;
+
+            if (!XrActivity.isEnabled()) {
+                getHolder().setSizeFromLayout();
+                return;
+            }
         }
 
         getDimensionsFromSettings();
--- a/app/src/main/java/com/termux/x11/XrActivity.java
+++ b/app/src/main/java/com/termux/x11/XrActivity.java
@@ -49,7 +49,6 @@ public class XrActivity extends MainActivity implements GLSurfaceView.Renderer {
 
     private boolean isImmersive = false;
     private boolean isSBS = false;
-    private boolean hasFocus = false;
     private boolean keyboardDisabled;
     private long lastEnter;
     private final float[] lastAxes = new float[ControllerAxis.values().length];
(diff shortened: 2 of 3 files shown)


app/src/main/java/com/termux/x11/XrActivity.java
```diff
@@ -183,8 +186,28 @@ public void onDrawFrame(GL10 gl10) {
...
     public boolean dispatchKeyEvent(KeyEvent event) {
+        // Ignore event madness when disabling software keyboard
+        if (keyboardDisabled) {
+            return true;
+        }
+
+        // Bluetooth/OTG keyboards work properly with the standard flow
+        if (hasFocus) {
+            return super.dispatchKeyEvent(event);
+        }
```
twaik, line 209: There is `hasWindowFocus` function, no need to store current window focus value in separate variable.

app/src/main/java/com/termux/x11/XrActivity.java
```diff
@@ -183,8 +186,28 @@ public void onDrawFrame(GL10 gl10) {
...
+        super.onWindowFocusChanged(hasFocus);
+        this.hasFocus = hasFocus;
+        if (hasFocus) {
+            keyboardDisabled = true;
+            new Handler().postDelayed(() -> keyboardDisabled = false, 250);
+        }
+    }
+
     @Override
     public boolean dispatchKeyEvent(KeyEvent event) {
```
twaik, line 200: Does `LorieView`'s dispatchKeyEventPreIme work in the case if Activity's dispatchKeyEvent is overridden?

twaik (changes requested):

app/src/main/java/com/termux/x11/XrActivity.java
lvonasek: Software keyboard sends randomly event with KEYCODE_MEDIA_PLAY into dispatchKeyEventPreIme.
Hardware keyboard works sends correct events into dispatchKeyEventPreIme.

twaik:
There are pending reviews.

app/src/main/java/com/termux/x11/XrActivity.java
lvonasek: I am sorry, I do not understand how is your question related to the code.

What change do you require here?

twaik:
@lvonasek can you please read pending reviews and respond?

app/src/main/java/com/termux/x11/LorieView.java
```diff
@@ -138,6 +138,10 @@ void getDimensionsFromSettings() {
         Prefs prefs = MainActivity.getPrefs();
         int width = getMeasuredWidth();
         int height = getMeasuredHeight();
+        if (XrActivity.isEnabled()) {
+            width = 1024;
+            height = 1024;
+        }
```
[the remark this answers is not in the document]
lvonasek: I made it 4:3 in https://github.com/termux/termux-x11/pull/671/commits/27b52aaafe26a25a84fd2218248cd7aa0e672024

app/src/main/java/com/termux/x11/XrActivity.java
lvonasek: I fixed that in https://github.com/termux/termux-x11/pull/671/commits/27b52aaafe26a25a84fd2218248cd7aa0e672024

lvonasek:
I responded to not answered reviews (although I thought this was clear, I am sorry for the misunderstanding).

About the answer from @natebragg:
> twaik that did not work, although I was only able to test it in XR mode, since the option to disable it was not present in the preferences

I think the user reacted to ctrl+space issue which isn't expected to work in the XR mode neither (the same issue like #665).

twaik:
They are still considered to be pending.
Screenshot_20240714_094501_Brave

lvonasek:
> They are still considered to be pending. Screenshot_20240714_094501_Brave

There must be some issue with GitHub. These comments aren't visible for me. But let me clarify it here:
`dispatchKeyEventPreIme` is called, with randomly receiving `KEYCODE_MEDIA_PLAY` from software keyboard. But only up events.

These events are not passing into `dispatchKeyEvents`. Do you still require to add `KEYCODE_MEDIA_PLAY` filtering although there is no negative effect by it?

I personally prefer to avoid "hacky" code if it is not necessary.

twaik:
What about the last one?

lvonasek:
> What about the last one?

Yes, `dispatchKeyEventPreIme` is called correctly.

Markadel1:
@lvonasek Is it possible to add support for other OpenXR-Compatible devices using the Monado runtime for Android? Monado has drivers for Nreal Air/Pro and Rokid glasses. I enjoy using OpenXR mode for Termux-X11 on the Quest, but being able to use it on my glasses would be great!

lvonasek:
* I do not support custom runtimes due to license madness.
* The second problem is that your headset doesn't support hybrid apps (switching between Android and VR runtime inside of one app).
* And without the headset it is hard to do anything.

PR #676, opened 2024-07-18
OpenXR - Switching between 2D and XR

lvonasek:
When user opens in the headset 2D preferences screen in XR mode then it behaves like two separate apps.

In this PR I did following changes:
* added a broadcast to close XR session from the 2D context
* modified the condition to support going from 2D into XR
* removed redundant `getMainDisplay`
* set native calls in `XrActivity` to `libxrio` private

### DIFF
--- a/app/src/main/java/com/termux/x11/MainActivity.java
+++ b/app/src/main/java/com/termux/x11/MainActivity.java
@@ -534,11 +534,17 @@ public class MainActivity extends AppCompatActivity implements View.OnApplyWindo
     void onPreferencesChangedCallback() {
         prefs.recheckStoringSecondaryDisplayPreferences();
 
-        if (oldXrMode != prefs.xrMode.get() && XrActivity.isSupported() &&
-                prefs.xrMode.get() != this instanceof XrActivity) {
-            /* Going back to 2d mode does not work */
-            // getBaseContext().startActivity(new Intent(this, MainActivity.class)
-            //        .addFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_CLEAR_TASK));
+        if (oldXrMode != prefs.xrMode.get() && XrActivity.isSupported()) {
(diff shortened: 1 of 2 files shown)


twaik:
No need to call `teardown` with broadcast. Simply make `protected void teardown() {}` in MainActivity`, make `teardown` protected function with `@Override` and it will be available from MainActivity...

lvonasek:
> No need to call `teardown` with broadcast. Simply make `protected void teardown() {}` in MainActivity`, make `teardown`protected function with`@Override` and it will be available from MainActivity...

I tried that, it didn't work.

twaik:
It dod not work with exception or it simply and silently ignored changes?

twaik:
I am not so sure about this but it seems to me you should pass it to GLSurfaceView's thread with `GLSurfaceView.queueEvent` because it is not async.

twaik:
And in clientConnectedStateChanged too.

lvonasek:
It simply and silently ignored changes.

As far as I understand it. When I open the preference screen from the system notifications, it opens a new instance of the app which doesn't contain XrActivity.

twaik:
That is correct, but why exactly you think it is related?

twaik:
BradcastReceiver receiving notification about preferences changed is registered in runtime. So XrActivity receives it too.

lvonasek:
It is like running two separate apps. The one with the preference screen doesn't contain instance of XrActivity.

twaik:
It is weird.
1. Only one activity should be available at the same time.
2. Broadcast about preferences change should be sent to both activities. It is being broadcasted. Something is broken.

lvonasek:
> 1. Only one activity should be available at the same time.

It isn't possible to mix XR and 2D content in one activity. As I start the XrActivity another activity is created. In my point of view the behavior is expected.

> 2. Broadcast about preferences change should be sent to both activities. It is being broadcasted. Something is broken.

I cannot say why the broadcast doesn't work. The XR behavior in Android apps isn't documented at all.

PR #711, opened 2024-08-29
to delete old workflows runs automate

ask9027:


Biswa96:
Why is this required? Old artifacts from CI help to test different versions without building.

ask9027:
artifacts will expired after some time then there is no use of them runs.

PR #768, opened 2024-12-11
enhancement(LorieView.java): improve input handling

twaik:
Making LorieView handle korean/chinese/turkish/other languages requiring composition and composed symbols.
@knyipab can you please test this? It replaces #620 solution.

@813ethan @ilhan-athn7 @robertkirkman @hansm629 can you please test this too?

### DIFF
--- a/app/src/main/cpp/lorie/activity.c
+++ b/app/src/main/cpp/lorie/activity.c
@@ -26,7 +26,7 @@ extern volatile int conn_fd; // The only variable from shared with X server code
 
 static struct {
     jclass self;
-    jmethodID getInstance, clientConnectedStateChanged;
+    jmethodID getInstance, clientConnectedStateChanged, resetIme;
 } MainActivity = {0};
 
 static struct {
@@ -119,6 +119,7 @@ static void nativeInit(JNIEnv *env, jobject thiz) {
         MainActivity.self = FindClassOrDie(env,  "com/termux/x11/MainActivity");
         MainActivity.getInstance = FindMethodOrDie(env, MainActivity.self, "getInstance", "()Lcom/termux/x11/MainActivity;", JNI_TRUE);
         MainActivity.clientConnectedStateChanged = FindMethodOrDie(env, MainActivity.self, "clientConnectedStateChanged", "()V", JNI_FALSE);
+        MainActivity.resetIme = FindMethodOrDie(env, (*env)->GetObjectClass(env, thiz), "resetIme", "()V", JNI_FALSE);
     }
 
     (*env)->GetJavaVM(env, &vm);
@@ -203,6 +204,9 @@ static int xcallback(int fd, int events, __unused void* data) {
                 case EVENT_REMOVE_BUFFER: {
                     rendererRemoveBuffer(e.removeBuffer.id);
                     break;
+                }
+                case EVENT_WINDOW_FOCUS_CHANGED: {
+                    (*env)->CallVoidMethod(env, thiz, MainActivity.resetIme);
                 }
             }
         }
@@ -292,6 +296,8 @@ static void sendWindowChange(__unused JNIEnv* env, __unused jobject cls, jint wi
 
 static void sendMouseEvent(__unused JNIEnv* env, __unused jobject cls, jfloat x, jfloat y, jint which_button, jboolean button_down, jboolean relative) {
     if (conn_fd != -1) {
+        if (which_button > 0)
+            (*env)->CallVoidMethod(env, globalThiz, MainActivity.resetIme);
         lorieEvent e = { .mouse = { .t = EVENT_MOUSE, .x = x, .y = y, .detail = which_button, .down = button_down, .relative = relative } };
         write(conn_fd, &e, sizeof(e));
     }
@@ -308,6 +314,7 @@ static void sendStylusEvent(__unused JNIEnv *env, __unused jobject thiz, jfloat 
                             jint pressure, jint tilt_x, jint tilt_y,
                             jint orientation, jint buttons, jboolean eraser, jboolean mouse) {
     if (conn_fd != -1) {
+        (*env)->CallVoidMethod(env, globalThiz, MainActivity.resetIme);
         lorieEvent e = { .stylus = { .t = EVENT_STYLUS, .x = x, .y = y, .pressure = pressure, .tilt_x = tilt_x, .tilt_y = tilt_y, .orientation = orientation, .buttons = buttons, .eraser = eraser, .mouse = mouse } };
         write(conn_fd, &e, sizeof(e));
     }
@@ -360,7 +367,7 @@ static void sendTextEvent(JNIEnv *env, __unused jobject thiz, jbyteArray text) {
             p += len;
             if (p - (char*) str >= length)
                 break;
-            usleep(30000);
+            usleep(2500);
         }
 
         (*env)->ReleaseByteArrayElements(env, text, str, JNI_ABORT);
@@ -391,7 +398,7 @@ JNIEXPORT jint JNI_OnLoad(JavaVM *vm, __unused void *reserved) {
             {"sendTouchEvent", "(IIII)V", (void *)&sendTouchEvent},
             {"sendStylusEvent", "(FFIIIIIZZ)V", (void *)&sendStylusEvent},
             {"requestStylusEnabled", "(Z)V", (void *)&requestStylusEnabled},
-            {"sendKeyEvent", "(IIZ)Z", (void *)&sendKeyEvent},
+            {"sendKeyEvent", "(IIZI)Z", (void *)&sendKeyEvent},
             {"sendTextEvent", "([B)V", (void *)&sendTextEvent},
             {"requestConnection", "()Z", (void *)&requestConnection},
     };
--- a/app/src/main/cpp/lorie/cmdentrypoint.c
+++ b/app/src/main/cpp/lorie/cmdentrypoint.c
@@ -472,6 +472,13 @@ void lorieUnregisterBuffer(LorieBuffer* buffer) {
     }
 }
 
+void DDXNotifyFocusChanged(void) {
+    if (conn_fd != -1) {
+        lorieEvent e = { .type = EVENT_WINDOW_FOCUS_CHANGED };
+        write(conn_fd, &e, sizeof(e));
+    }
+}
+
 JNIEXPORT jobject JNICALL
 Java_com_termux_x11_CmdEntryPoint_getXConnection(JNIEnv *env, __unused jobject cls) {
     int client[2];
--- a/app/src/main/cpp/lorie/lorie.h
+++ b/app/src/main/cpp/lorie/lorie.h
@@ -102,6 +102,7 @@ typedef enum {
     EVENT_CLIPBOARD_ANNOUNCE,
     EVENT_CLIPBOARD_REQUEST,
     EVENT_CLIPBOARD_SEND,
+    EVENT_WINDOW_FOCUS_CHANGED,
 } eventType;
 
 typedef union {
--- a/app/src/main/cpp/patches/xserver.patch
+++ b/app/src/main/cpp/patches/xserver.patch
@@ -328,4 +328,23 @@ index f9b7b06d9..f4b2aeddc 100644
 +                screen_priv->info->after_flip(vblank->crtc, vblank->event_id, ust, vblank->target_msc, vblank->pixmap);
                  return;
              }
++++ b/dix/enterleave.c
+@@ -1540,6 +1540,8 @@ DeviceFocusEvents(DeviceIntPtr dev, WindowPtr from, WindowPtr to, int mode)
+     }
+ }
+ 
++extern void DDXNotifyFocusChanged(void);
++
+ /**
+  * Figure out if focus events are necessary and send them to the
+  * appropriate windows.
+@@ -1550,6 +1552,9 @@ DeviceFocusEvents(DeviceIntPtr dev, WindowPtr from, WindowPtr to, int mode)
+ void
+ DoFocusEvents(DeviceIntPtr pDev, WindowPtr from, WindowPtr to, int mode)
+ {
++    if (from != to)
++        DDXNotifyFocusChanged();
++
+     if (!IsKeyboardDevice(pDev))
+         return;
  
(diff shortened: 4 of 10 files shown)


twaik:
The solution is created for software keyboard, I did not try to fix hardware keyboard.
Also in the case if you use "prefer scancodes when possible" input will be handled by fcitx or similar mechanism in linux, not by my code.

robertkirkman:
In Android Studio if I click the Debug button using this PR, then reproduce the crash, the debugger does not seem to print any backtrace, but this appears in `adb logcat`, when I type any character in Chromium inside a Termux:X11 app built from this PR using Gboard, and I can confirm that with this PR I am able to type "Ç" but Backspace makes the app close, but without this PR I cannot type "Ç" but backspace does not make the app close.

<details>

<summary>logcat snippet</summary>

```
2024-12-11 08:24:05.312 17579-17579 DEVICES                 com.termux.x11                       D  found device "hall" sources 0x80000000
2024-12-11 08:24:05.312 17579-17579 DEVICES                 com.termux.x11                       D  requesting stylus false
2024-12-11 08:24:05.312 17579-17579 DEVICES                 com.termux.x11                       D  external keyboard connected false
2024-12-11 08:24:05.345 17579-17579 ViewRootIm...nActivity] com.termux.x11                       I  Relayout returned: old=(0,0,1440,2960) new=(0,0,1440,2960) req=(1440,2960)0 dur=16 res=0x1 s={true 538304110592} ch=false
2024-12-11 08:24:10.778 17579-17579 LorieNative             com.termux.x11                       D  Parsing text: æ
2024-12-11 08:24:10.778 17579-17579 LorieNative             com.termux.x11                       D  Sending unicode event: æ (U+E6)
2024-12-11 08:24:12.181 17579-17579 AndroidRuntime          com.termux.x11                       D  Shutting down VM
2024-12-11 08:24:12.203 17579-17579 AndroidRuntime          com.termux.x11                       E  FATAL EXCEPTION: main
                                                                                                    Process: com.termux.x11, PID: 17579
                                                                                                    java.lang.IndexOutOfBoundsException: replace (0 ... -1) has end before start
                                                                                                    	at android.text.SpannableStringBuilder.checkRange(SpannableStringBuilder.java:1328)
                                                                                                    	at android.text.SpannableStringBuilder.replace(SpannableStringBuilder.java:523)
                                                                                                    	at android.text.SpannableStringBuilder.delete(SpannableStringBuilder.java:231)
                                                                                                    	at android.text.SpannableStringBuilder.delete(SpannableStringBuilder.java:38)
                                                                                                    	at com.termux.x11.LorieView$1.delete(LorieView.java:82)
                                                                                                    	at android.view.inputmethod.BaseInputConnection.deleteSurroundingText(BaseInputConnection.java:252)
                                                                                                    	at com.android.internal.view.IInputConnectionWrapper.executeMessage(IInputConnectionWrapper.java:450)
                                                                                                    	at com.android.internal.view.IInputConnectionWrapper$MyHandler.handleMessage(IInputConnectionWrapper.java:89)
                                                                                                    	at android.os.Handler.dispatchMessage(Handler.java:107)
                                                                                                    	at android.os.Looper.loop(Looper.java:237)
                                                                                                    	at android.app.ActivityThread.main(ActivityThread.java:8167)
                                                                                                    	at java.lang.reflect.Method.invoke(Native Method)
                                                                                                    	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:496)
                                                                                                    	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1100)
2024-12-11 08:24:12.302 17579-17579 Process                 com.termux.x11                       I  Sending signal. PID: 17579 SIG: 9
2024-12-11 08:24:12.403  1322-1841  InputDispatcher         pid-1322                             E  channel '8bb50eb com.termux.x11/com.termux.x11.MainActivity (server)' ~ Channel is unrecoverably broken and will be disposed!
2024-12-11 08:24:12.465  1322-3564  WindowManager           pid-1322                             E  win=Window{8bb50eb u0 com.termux.x11/com.termux.x11.MainActivity EXITING} destroySurfaces: appStopped=false win.mWindowRemovalAllowed=true win.mRemoveOnExit=true win.mViewVisibility=0 caller=com.android.server.wm.AppWindowToken.destroySurfaces:1249 com.android.server.wm.AppWindowToken.destroySurfaces:1230 com.android.server.wm.WindowState.onExitAnimationDone:5189 com.android.server.wm.-$$Lambda$01bPtngJg5AqEoOWfW3rWfV7MH4.accept:2 java.util.ArrayList.forEach:1262 com.android.server.wm.AppWindowToken.onAnimationFinished:3941 com.android.server.wm.AppWindowToken.commitVisibility:914 
2024-12-11 08:24:12.917   904-999   BufferQueueProducer     pid-904                              E  [SurfaceView - com.termux.x11/com.termux.x11.MainActivity@23442d7@9#0] dequeueBuffer: BufferQueue has been abandoned

```

</details>

twaik:
> Does that mean testing should be conducted using the Android software keyboard instead of a USB or Bluetooth hardware keyboard?

Software keyboard only.

> I can confirm that with this PR I am able to type "Ç" but Backspace makes the app close

Can you please try again?

twaik:
@robertkirkman And what software keyboard layout did you use?

twaik:
I mean I used gboard and it was fine on my device.

twaik:
It lets us use cursor control, keyboard suggestions (with word replacings!) and some other stuff.
https://github.com/user-attachments/assets/a62a0d5e-fd3c-41ec-b22b-0c1fa07f3fa2

twaik:
Korean input seems to be a bit more weird.

https://github.com/user-attachments/assets/b2726b01-fb4e-47c8-b43d-b67b556f08bd

hansm629:
@twaik 
Should I also test this test with the option off?

`Hardware keyboard scancodes workaround`


https://github.com/user-attachments/assets/f627e034-a8c3-451d-b4f7-24544ba5df96

ilhan-athn7:
> @813ethan @ilhan-athn7 @robertkirkman @hansm629 can you please test this too?

The cursor movement isn't usable on LatinIME
But this one works flawlessly on my end.

twaik:
> The cursor movement isn't usable on LatinIME

I did not find cursor control feature there. Probably it is not available there.

ilhan-athn7:
> > The cursor movement isn't usable on LatinIME
> 
> I did not find cursor control feature there. Probably it is not available there.


https://github.com/user-attachments/assets/5fe10af0-f397-442b-aabb-fedb137ef19f

twaik:
It is not the same IME I have. 
photo_2024-12-12_19-40-09

813ethan:
seems fine with gboard chinese handwriting and cantonese input for now

logs like this appeared while typing tho
```
(II) Added unknown keysym 0x10089ba to keycode 202
(II) Added unknown keysym 0x1005f97 to keycode 184
(II) Added unknown keysym 0x1005f88 to keycode 183
```

813ethan:
the `Deactivate special keys on additional key bar after each key press` option is also not working on this pr

twaik:
I think I need more data to reproduce "slow typing" and "autoccomplete". I need output of `termux-x11-preference`, apk/link to google play/f-droid of IME you use, name of layout, other factors that probably will help me.

twaik:
> the `Deactivate special keys on additional key bar after each key press` option is also not working on this pr

Should be fixed now.

> The cursor movement isn't usable on LatinIME

Should be fixed too but I am not sure about this.

813ethan:
nice, seems fine for now with gboard

ilhan-athn7:
> > The cursor movement isn't usable on LatinIME
>
> Should be fixed too but I am not sure about this.

It doesn't, it also brings back the previuos bug, where i can't use Backspace key and few letter

twaik:
Probably I should check what happens with your IME. Can you send the apk of your keyboard?

twaik:
@hansm629 I checked korean input with external keyboard and now I understand what you mean by saying "slow input". That is because of symbol composition, pending symbol was displayed at the bottom of screen. But it seems like I found some simple and pretty much dumb way to avoid using the composition displaying. Check the latest artifact.

twaik:
@813ethan probably I need your confirmation too because I checked only pinyin 3x4 of samsung keyboard, I am not sure if it will work on other IMEs.

813ethan:
seems working fine

https://github.com/user-attachments/assets/9c4dd4b3-a2ad-46e3-b126-c91b6d9983b9

twaik:
Lol. Current solution simply erases old text (with backspace key) Android thinks we compose and replaces it with new text. Combined with input delay it shows interesting result.

https://github.com/user-attachments/assets/ed8fe8fc-aa5c-4fef-b03e-22fe0876b249

813ethan:
also autocomplete is not always working?
https://github.com/user-attachments/assets/4ac127f7-a03b-46c7-9d6d-d310c2dfc80a

twaik:
Yeah, seems so. I am already on it. Is it gboard?

813ethan:
yes gboard

twaik:
This build should fix this.

twaik:
> Have you ever try to use backspace in Gboard English and other non-CJK langauges (with autocomplete and word suggestion)? The IME incorrectly think that it is working on some composed text that is yet to be commited, making backspace not functioning well.

Seems to be fixed now.

813ethan:
bug seems to be still happening

ilhan-athn7:
> Probably I'll disable predictive typing and word suggestion completely since it does not seem to be possible to make it work fine even on most common keyboards in our case (without direct access to ibus/fcitx/xim servers).

this scenarious can be handled by adding a switch in preferences, not sure if it worth the effort

twaik:
It is not worth the effort to make it work.

twaik:
Ok, seems like there is no way to disable text prediction and suggestions except using `InputType.TYPE_NULL` or old, "dumb" keyboard option, which will block people from using CJK languages... Sad. Probably I should implement some fake & dumb cursor tracking to make it work fine.

twaik:
> * when using gboard English IME with physical keyboard, Ctrl + Backspace deletes more words than expected which can be annoying. May be related to the current approach to compose text.

It is so dumb... I found it out. When you do CTRL+Backspace Android sends both Press and Release events for CTRL, but only Press release for Backspace, forgetting to send backspace releasing event. I've found a workaround, starting delayed sending key release event in `dispatchKeyEventPreIme` (before IME consumes it) and cancelling it in `dispatchKeyEvent` in the case if actually did not consume it...

> when using gboard Chinese IME + English IME with physical keyboard (with Gboard option "Physical keyboard" -> "Show on-screen keyboard" off), pressing SHIFT key will stuck the SHIFT key in Termux:X11.

I think same applies here.

twaik:
Ok, I hope I did not break anything. Also I added a switch to disable both autosuggestions/predictive text and CJK languages (because I can not disable them separately). Should work fine but I am not so sure about this.

twaik:
I need some feedback to decide if I merge it or not. I need someone to test it. 
CJK and autocomplete are suppressed by default, it can be enabled in preferences.

mexsrc:
( @twaik )
Look like no one just response it's test video to those pull request respectfully. I just downloaded ``.apk`` and ``.zip`` from check-in of those pull request and love to treat myself as tester.

It's *reasonable* because treating uneasy someone as tester is unsorted and can become harder. I'll must first install desktop environment with ``ensurepip`` variant of ``pipewire`` precisely, must secondly switch **Termux X11** from [ewt4's fork](https://github.com/ewt45/termux-x11-fork) to your work's application add-on and will thirsty test out for determining if I have such issue.

robertkirkman:
This seems good to me, in most programs, this makes the Backspace work.

However, in Luanti, the Backspace still does not work while using this.
The behavior of the backspace not working in Luanti persists consistently:
- Across all text fields in Luanti GUI
- Across multiple keyboard apps
- Across both hardware and software keyboards
- Regardless of whether "Enforce char based input" is enabled or disabled
- Regardless of whether "Prefer scancodes when possible" is enabled or disabled
- Regardless of whether "Hardware keyboard scancodes workaround" is enabled or disabled
- Regardless of whether all instances of `setlocale(..., "")` in Luanti code are deleted and the package recompiled

However, Luanti **does** receive an Event and Scancode of something, when the Backspace is touched.
When I apply this change in the Luanti code and recompile it, the Backspace works in Luanti for all combinations of the above configurations as long as this PR of Termux:X11 is used.

```diff
--- a/irr/include/Keycodes.h
+++ b/irr/include/Keycodes.h
@@ -16,7 +16,7 @@ enum EKEY_CODE
 	KEY_MBUTTON = 0x04,           // Middle mouse button (three-button mouse)
 	KEY_XBUTTON1 = 0x05,          // Windows 2000/XP: X1 mouse button
 	KEY_XBUTTON2 = 0x06,          // Windows 2000/XP: X2 mouse button
-	KEY_BACK = 0x08,              // BACKSPACE key
+	KEY_BACK = 0x9E,              // BACKSPACE key
 	KEY_TAB = 0x09,               // TAB key
 	KEY_CLEAR = 0x0C,             // CLEAR key
 	KEY_RETURN = 0x0D,            // ENTER key
```
This value change comes from stepping in a debugger and observing the functions where the Backspace is failing to be fully detected, then editing a value in such a way that forces the Backspace event to be detected.

I really do not understand currently why this is necessary only for Luanti and not other programs, but I have not been able to find any working solution yet that does not involve editing the value of a keycode in some part of Luanti code.

It would appear though that, if this PR enters the Termux:X11 repository, then I guess I could send this patch into the termux-packages/x11-packages/luanti folder and maybe that would be sufficient to connect backspace to Luanti. I'm not sure, it seems like that is a very messy thing to do without finding the root cause.

robertkirkman:
When I test some more, I can see that the backspace problem in luanti is technically something caused by the luanti package itself, and not entirely by Termux:X11, because when Termux:X11 is not used and instead a different X11 server implementation like `$PREFIX/bin/Xvnc` from the `tigervnc` package is used, the problem is still reproducible there, and also, the same value change inside the `Keycodes.h` file of luanti's code also works around the problem successfully and causes backspace to work there too. However, it is not technically a problem with upstream luanti either because luanti builds for GNU/Linux are not affected, regardless of whether they were built in SDL2 mode or not.

Because of that, if I am unable to create a better solution, when the time comes I should add that patch into the luanti package, along with changes for any other keycode problems I am able to find,
but I will still wait and synchronize that with this PR, because the game is practically the most usable with several graphics drivers that are only for Termux:X11, so a patch like that would be the most useful after Termux:X11 itself has backspace support in general, for all other programs.

twaik:
Ok, I do not see any problems here, so let's give it a try.

PR #810, opened 2025-02-20
Extra features for Termux-X11

moio9:
I recently forked Termux-X11 and added extra options, such as virtual screen key mapping and gamepad support. Would you consider merging these changes, as the project doesn’t seem to be very game-oriented? If not, is there anything I could improve?


twaik (changes requested):
I am not intended nor to change app name, nor to update readme with this new name, nor to change app style, nor to change indentation of already present code. Also commit descriptions are unclear and not informative, you should squash commits or fix descriptions.

`com.samsung.android.multidisplay.keep_process_alive` is not needed since activity does not do any background job. It is a dumb client of X server which runs in `com.termux` app sandbox. 
Comments in language other than English are not acceptable.
And it is not the whole list, I only scrolled through this.

moio9:
I just wanted to know if you were interested in any of the features before I started polishing or removing them. That's why I asked if there was anything to improve. I wasn't talking about things like the readme or the name; those are just for my fork so that people can find it more easily.

Issue #1022, opened 2026-05-11
[Feature]: Experimental workaround for Samsung OneUI 8.x cpuset/performance throttling with Termux-X11


hansm629:
### Problem description

@twaik 
Hello. sir
https://github.com/termux/termux-app/issues/5086

Currently on Samsung OneUI 8.x devices, when the Termux app is in the background, there is a symptom where CPU performance is degraded and the big core cluster is not utilized at all.

I would like to share an experimental workaround for a performance issue I observed on Samsung OneUI 8.x when using Termux:X11 in DeX mode.

This is not something I am immediately comfortable submitting as a PR to the main branch, because it involves `sharedUserId`, process spawning from the Android Activity, and foreground service lifecycle handling. I would like to ask for maintainer feedback first.

### What steps will reproduce the bug?

```
## Test environment

- Device: Samsung OneUI 8.x device
- Android: Android 16 / OneUI 8.x
- SoC/GPU: Qualcomm / Adreno 840
- Termux app: GitHub debug build, v0.119.0-beta.3
- Termux:X11: custom build based on latest source
- Desktop: XFCE4
- Test workload: Shotcut 4K video rendering, Chromium, general XFCE apps

## Problem

In the normal Termux:X11 + XFCE4 setup, the process tree is roughly:

```text
Termux shell
 └─ termux-x11 :0
 └─ xfce4-session
     └─ xfce4-panel
         └─ shotcut
             └─ melt-7
``` 

The Android com.termux.x11 app itself becomes the visible foreground/top-app process, but the actual native workloads such as xfce4-session, shotcut, and melt-7 are spawned from the Termux shell / XFCE side.

On my device, these native workloads did not inherit the /top-app cpuset.

For example, during Shotcut rendering:

```
shotcut
/foreground
Cpus_allowed_list: 0-5

melt-7
/foreground
Cpus_allowed_list: 0-5
``` 
This is an 8-core device, so CPU 6-7, the big/prime cores, were not fully available to the actual rendering workload.

In practice, Shotcut 4K rendering initially showed about 10 minutes ETA, but under thermal throttling it could increase to 15+ minutes.

# First workaround: sharedUserId
I first added this to the Termux:X11 manifest: (https://github.com/hansm629/termux-x11/actions/runs/25434788182)

```
android:sharedUserId="com.termux"
``` 
This successfully made Termux and Termux:X11 share the same UID:
```
com.termux      uid: 10400
com.termux.x11  uid: 10400
``` 
[...]

PR #1023, opened 2026-05-11
Add experimental overlay mode for Termux:X11

### LINKED ISSUE #442: Overlayed Termux-x11
I recently conducted the following test: I launched a banal benchmark of gears as usual (I went to termux, then to termux x11), and then I tried to run all this in picture-in-picture mode (instead of termux-x11 I switched to termux). Fps itself grow by as much as 40-80%.I concluded: on my Android firmware, Termux get fewer resources in the background than it can. This also explains why my sound lags when I'm in termux-x11, but stops lagging as soon as I switch to termux.I would really like the author of termux-x11 to do something like virgl overlay. Then android will assume that we are in termux itself and we will get from this 
1) Maximum performance 
2) Normal sound (specifically in my case)
So is this possible? PiP mode is not what I need because I can't stretch it on full screen + I can't control mouse/other things in PiP mode.

twaik:
Move activity content into TYPE_APPLICATION_OVERLAY window with movable/resizable overlay decorations and icon minimize support.

Mitigates #1022
Fixes #442

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -3,6 +3,8 @@
     xmlns:tools="http://schemas.android.com/tools"
     android:installLocation="internalOnly">
 
+    <uses-permission android:name="com.termux.permission.RUN_COMMAND" />
+    <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.WRITE_SECURE_SETTINGS" tools:ignore="ProtectedPermissions" />
--- a/app/src/main/java/com/termux/x11/LoriePreferences.java
+++ b/app/src/main/java/com/termux/x11/LoriePreferences.java
@@ -7,6 +7,8 @@ import static android.content.pm.PackageManager.PERMISSION_GRANTED;
 import static android.os.Build.VERSION.SDK_INT;
 import static android.system.Os.getuid;
 
+import static androidx.core.content.ContentProviderCompat.requireContext;
+
 import android.annotation.SuppressLint;
 import android.content.BroadcastReceiver;
 import android.content.ClipData;
@@ -187,6 +189,7 @@ public class LoriePreferences extends AppCompatActivity implements PreferenceFra
     }
 
     public static class LoriePreferenceFragment extends PreferenceFragmentCompat implements OnPreferenceChangeListener {
+        private static boolean mPendingOverlayPermissionRequest = false;
         private final Runnable updateLayout = this::updatePreferencesLayout;
         private static final Method onSetInitialValue;
         static {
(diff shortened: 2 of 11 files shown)


twaik:
@hansm629 @robertkirkman it is kinda buggy and I am aware about the bug with reparenting to main activity after "closing" and that it does not remember its position and size but I kinda need some review.
I am pretty much sure it should mitigate the cpu/gpu throttling problem.

twaik:
it does not check for overlay permission yet, you should give it manually.

twaik:
Few bugs fixed, maybe few more added.
Need more reviews.

robertkirkman:
@RalfWerner I am not sure if this would completely be useful to you, but this would allow another kind of what you call "pop view", would you want to try it?

hansm629:
@twaik 
I apologize for the late reply.
I've been a bit busy with work.

Just then, my Galaxy S26 Ultra was updated to the `OneUI 9 Beta (Android 17)`.

I will test it in that environment.

hansm629:
@twaik 
Here is a report on the test results.

We confirmed that when running in overlay mode, all CPU cores are utilized under heavy workloads.

However, there are a few issues.
Screenshot_20260513_212056_Termux




# 1. Overlay Full-Screen Mode

Is there currently no way to operate the overlay in full-screen mode?

In DeX mode, unfortunately, I was unable to increase the window size beyond the size shown in the screenshot attached above.

# 2. Overlay Window Display Only When X Server Is Running

Currently, the overlay window is displayed indiscriminately, and it remains even after using the command to close the X Server and the Termux-X11 app.

Is it possible to implement a system where the overlay window is displayed only when the X Server execution command is used, and closes when the command to close the X Server and the Termux-X11 app is entered?

# 3. Keyboard Initiation in Device On-Screen Mode

When using the device on-screen mode, the software keyboard is not being invoked via the Android back key. Is it possible to improve this aspect?

robertkirkman:
> 3. Keyboard Initiation in Device On-Screen Mode

> When using the device on-screen mode, the software keyboard is not being invoked via the Android back key. Is it possible to improve this aspect?

I neglected to mention this here yet, but there is a global solution to this problem for **all failures of Android keyboard to open all in all apps Termux or non-Termux alike**.

It is the "global notification" setting of the org.pocketworkstation.pckeyboard app.

https://f-droid.org/en/packages/org.pocketworkstation.pckeyboard/

image

twaik:
Closing because new sharedUserId (d24418e4fdc2a84313048d787430d1d69ad6dab8) lets you use termux without going to /foreground cpuset, stays /top-app as long as termux-app or termux-x11 is in foreground and focused.

robertkirkman:
> > 3. Keyboard Initiation in Device On-Screen Mode
> 
> > When using the device on-screen mode, the software keyboard is not being invoked via the Android back key. Is it possible to improve this aspect?
> 
> I neglected to mention this here yet, but there is a global solution to this problem for **all failures of Android keyboard to open all in all apps Termux or non-Termux alike**.
> 
> It is the "global notification" setting of the org.pocketworkstation.pckeyboard app.
> 
> https://f-droid.org/en/packages/org.pocketworkstation.pckeyboard/
> image

I forgot to explain that in One UI 8 (pictured) it's easy to open that notification but in LineageOS 23 it takes approximately 200 tries in a row to open that notification but eventually it does work even though it seems to everyone who doesn't try about 200 times in a row that it doesn't work, but that's ridiculous so it's good that there is now a sharedUserID version that can prevent throttling while in fullscreen and have the back button to open the keyboard working.

https://www.youtube.com/watch?v=yu8BSAdCkts&t=1027s

PR #1024, opened 2026-05-12
feat(displaylink): implement native DisplayLink USB display sink

XiaoTong6666:
Introduces a complete hardware-accelerated rendering pipeline and USB bridge to output the Termux-X11 display directly to DisplayLink monitors and docks. 

- Driver Injection: Dynamically loads `libDisplayLinkManager.so` and extracts `.spkg` firmware from the official `com.displaylink.presenter` app, bypassing the need to bundle proprietary binaries.
- Native Render Sink: Intercepts the X11 drawing pipeline in `renderer.c`. Creates an offscreen FBO to render the root window and composite the hardware cursor, automatically scaling and letterboxing the output to match the connected monitor's preferred resolution.
- Zero-Copy JNI Pipeline: Utilizes `glReadPixels` to read raw RGBA buffers from the GL thread, passing them to the Java bridge via `DirectByteBuffer` for DisplayLink hardware encoding.
- USB Host Management: Adds `DisplayLinkBridge` to handle device hotplugging, DDC monitor info parsing, dynamic permission requests, and watchdog-based driver recovery.
- Diagnostics: Includes `DisplayLinkDebugReceiver` for adb-driven lifecycle control and marker-based PPM frame dumping for rendering isolation tests.

Note: Requires the official DisplayLink Presenter app to be installed on the host device.

### DIFF
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -17,6 +17,22 @@
     public static void main(java.lang.String[]);
 }
 
+# Keep DisplayLink bridge and all JNI-reflected classes/members.
+-keep class com.termux.x11.displaylink.DisplayLinkBridge {
+    public static boolean pushFrame(java.nio.ByteBuffer, int, int, int);
+    public static com.termux.x11.displaylink.DisplayLinkBridge getInstance();
+    public void start(android.content.Context);
+    public void stop();
+    public void onDisplayConnected(long);
+    public void onDisplayDisconnected(long);
+    public void onMonitorInfo(long, com.displaylink.manager.display.MonitorInfo);
+}
+
+-keep class com.displaylink.manager.NativeDriver { *; }
+-keep class com.displaylink.manager.NativeDriverListener { *; }
+-keep class com.displaylink.manager.display.DisplayMode { *; }
+-keep class com.displaylink.manager.display.MonitorInfo { *; }
+
 # Keep Preference subclasses' onSetInitialValue for reflective calls
 -keepclassmembers class * extends android.preference.Preference {
     void onSetInitialValue(boolean, java.lang.Object);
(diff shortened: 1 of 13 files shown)


XiaoTong6666:
Disclaimer:
This commit does not contain, distribute, or bundle any proprietary code or firmware from DisplayLink (Synaptics). It implements a dynamic JNI bridge for interoperability purposes. Users must independently install the official DisplayLink Presenter application for this feature to function. This integration utilizes undocumented APIs via dynamic reflection, is strictly unofficial, and is provided without any commercial support or warranty.

twaik:
1. Seems like a result of reverse engineering of proprietary software which may violate its EULA/ToS.
2. It has the same problem as the regular Presentation Mode: main screen of device is required to be kept on, the Termux:X11 activity must be on screen and focused.
3. You already can acquire ImageReader from native library specifically for writing images there. Why exactly do you need to reimplement pushing buffers if you can simply use ImageReader's surface/anativewindow as a rendering target???

XiaoTong6666:
Haha, you're totally right on all points!

Honestly, this is just a quick PoC right now. I went with the push-buffer + glReadPixels approach simply because it was the fastest/laziest way to validate if the DisplayLink stack would even accept our frames, without doing a deep surgery on the rendering pipeline first.

You're spot on about the ImageReader and Surface. Switching to a proper Surface rendering target is definitely the right way to do this, and theoretically, it should finally bypass that annoying foreground/screen-on limitation you mentioned.

Regarding the EULA/ToS aspect, the patch is designed to stay clean on our end—it doesn't bundle or distribute any proprietary blobs, and purely relies on dynamic reflection for interoperability with the user's independently installed app.

I'll drop the push-buffer workaround and rework the prototype to directly use the Surface path. Will update the PR once I get it working properly!

twaik:
1. I am talking to pr author. Not to ai. 
2. Switching rendering to imagereader's surface will not drop screen-on limitation.
3. Not bundling binary blobs does not drop the fact it is reverse engineering of proprietary product.

XiaoTong6666:
my bad. I am not a native English speaker, so I used an AI to translate and polish my thoughts, and it clearly came out sounding like a bot.

you are absolutely right on the technical side. the activity lifecycle will still pause the render loop when backgrounded/screen off, so just swapping the surface target wouldn't magically fix the presentation mode issue.

and I completely understand your policy stance on the reverse engineering part. since relying on these undocumented APIs is a hard blocker for the project, I don't want to cause any legal headaches for the repo.

I will close this PR and just maintain it as a patch in my own fork for anyone who needs it. thanks for taking the time to review the code anyway.

Issue #904, opened 2025-08-22
[Bug]: Mod4/Meta key not releasing on next key press with i3


jorexdeveloper:
### Problem description

Starting with...

```bash
TERMUX_X11_XSTARTUP="dbus-launch --exit-with-session i3"
termux-x11 :0
```

Launches termux-x11 as expected;

# Problem

When i press `Meta` key, then `Enter` key to lauch a terminal

**The `Meta` key remains active!**

Such that any other key pressed resolves to `meta+<that key>`...which is not what should happen

This continues forever...until Meta key is pressed again...Then the outcome is undefined; that is -- sometimes the meta key is released and sometimes it isn't...

Using latest release **1.03.01-2a1fac2-09.06.25** downloaded from releases page on github...

# What i have tried

Enabling accessibility...(tho WRITE_SECURE_SETTINGS not granted)

trying all possible permutations of keyboard settings

Extra info...(probably not relevant)

Keyboard - Unexpected Keyboard (from fdroid/github

### What steps will reproduce the bug?

1. Start i3 as explained above...
2. Meta+Enter (or any key + Meta)

### What is the expected behavior?

Meta4+Enter should release the meta key...

twaik:
Try the debug artifact from here.
https://github.com/termux/termux-x11/pull/1025

robertkirkman:
I am able to reproduce this bug, and for me, the PR does change the behavior to be very slightly better, I think, but unfortunately, for me it does not seem to completely fix the problem

PR #1025, opened 2026-05-12
fix: release stale modifiers in accessibility service

### LINKED ISSUE #781: [Bug]: mouse click has no effect after switching between android apps via alt-tab
### Problem description

When I leave termux-x11 for another android app through keyboard alt-tab combination on mi Redmi Pad Pro tablet, once I go back to Termux-X11, all seems to be working but the mouse clicks have no effect.
When I leave termux-X11 through gestures (rather than alt-tab), all keep working when returning to termux-X11.

install proot-distro ubuntu + xfce4
start xfce4 from termux using the script below:
```

#!/data/data/com.termux/files/usr/bin/bash

# Kill open X11 processes
kill -9 $(pgrep -f "termux.x11") 2>/dev/null

# Enable PulseAudio over Network
pulseaudio --start --load="module-native-protocol-tcp auth-ip-acl=127.0.0.1 auth-anonymous=1" --exit-idle-time=-1

# Prepare termux-x11 session
#export XDG_RUNTIME_DIR=${TMPDIR}
export DISPLAY=:0
termux-x11 :0 &

# Wait a bit until termux-x11 gets started.
sleep 3

# Launch Termux X11 main activity
am start --user 0 -n com.termux.x11/com.termux.x11.MainActivity
sleep 1

# Login in PRoot Environment. Do some initialization for PulseAudio, /tmp directory
# and run KDE as user droidmaster.
# See also: https://github.com/termux/proot-distro
# Argument -- acts as terminator of proot-distro login options processing.
# All arguments behind it would not be treated as options of PRoot Distro.
proot-distro login ubuntu --shared-tmp -- /bin/bash -c  'export PULSE_SERVER=127.0.0.1 && export DISPLAY=:0 && export GALIUM_DRIVER=virpipe && export BROWSER=firefox && dbus-launch -
-exit-with-session startxfce4'

exit 0
```

Switch to other android apps through alt-tab

return to termux-X11

Try to open a menu or an app -> no effect when you click

click should still have effects when returning to the termux-X11 app

twaik:


twaik:
@robertkirkman @Digit-al @berilevi @nathaneltitane @jorexdeveloper
can you please check if the problem resolved in this build?

robertkirkman:
I will test it

robertkirkman:
Ok, I have tested it very carefully, and unless I am making a mistake, the problem in `i3` (reported by jorexdeveloper in 2025 and then reproduced independently by me based on their directions)  is unfortunately still reproducible for me,

if this is not reproducible on your device, should I try to take video recordings or allow you to connect with scrcpy?

I should note this: it does seem to be difficult to reproduce in terms of inputs. The inputs required to reproduce it kind of involve "mashing", i.e. a basic QA tester's technique of pushing the button rapidly in a specific pattern with subtle variations in timing that I would say "feels right to reproduce it".

Once a Termux:X11 session has entered the bugged state, it is possible to persistently reproduce the problem without repeating the "mashing technique" until the Termux:X11 is force quit or the meta key is pushed again, and I have one currently in the bugged state which I can use to perform further tests if needed.

twaik:
What about the alt-tab-related bug?

robertkirkman:
> What about the alt-tab-related bug?

I had not seen that before. 

In order to be able to reproduce that bug (in the master branch), I first had to use this command:

```bash
termux-x11-preference "dexMetaKeyCapture":"false"
```

It seems that I had that setting "true", which might have prevented me from reproducing it.

After that, I was able to reproduce it in the master branch version of Termux:X11, but not in this PR version, so that does seem to be fixed for me in this.

twaik:
Does it cause any other problems?

robertkirkman:
As far as I noticed so far, it does not cause other problems, and it may make https://github.com/termux/termux-x11/issues/904 slightly more difficult to reproduce, but for me does not make https://github.com/termux/termux-x11/issues/904 completely impossible to reproduce

robertkirkman:
I am not sure if it makes #781 completely impossible to reproduce, but probably it at least makes it harder to reproduce, since I could not yet while using this PR but I could easily in the master branch.

PR #1049, opened 2026-07-05
fix(input): filter spurious key events from touchscreen hardware

DdogezD:
## Problem

On some devices with Oplus/OnePlus touchpanel firmware, the kernel module `oplus_keyevent_notifier` registers the touchscreen as a keyboard input source. During touch and mouse operations the touchpanel injects spurious function key events (e.g. F4 on Android keycode 134) into the input pipeline.

These events reach the X server as real key presses, causing unexpected modifier/function key behavior in X11 applications. For example, `^J` (Justify) being triggered in nano when moving the mouse.

## Root cause

The touchpanel hardware (`device=touchpanel1`) reports events with `source=0x101` (SOURCE_KEYBOARD) and valid scancodes. Termux:X11 forwards all keyboard events unconditionally. There is no mechanism to distinguish legitimate keyboard input from touchscreen-generated key events.

## Fix

Add a boolean preference `filterTouchscreenKeyEvents` (enabled by default) in the Keyboard settings screen. When enabled, `InputEventSender.sendKeyEvent()` checks if the event's device name contains "touchpanel" and silently discards the event before any further processing.

This is safe because:
- Real keyboards have device names like "Virtual", "bluetooth keyboard", etc.
- IME text input goes through `commitText()`, not `sendKeyEvent()`
- The filter only affects the `EVENT_KEY` path, not text or IME input
- Users with unaffected devices can disable the filter in settings

## Testing

Built and tested on an Oplus device (Android 15 / ARM64) with chroot XFCE via Termux:X11. After the fix, spurious F4 key events from touchpanel1 are filtered and no longer reach the X server. Normal keyboard input and IME text input continue to work correctly.

### DIFF
--- a/app/src/main/java/com/termux/x11/input/InputEventSender.java
+++ b/app/src/main/java/com/termux/x11/input/InputEventSender.java
@@ -40,6 +40,7 @@ public final class InputEventSender {
     public boolean pauseKeyInterceptingWithEsc = false;
     public boolean stylusIsMouse = false;
     public boolean stylusButtonContactModifierMode = false;
+    public boolean filterTouchscreenKeyEvents = true;
 
     /** Set of pressed keys for which we've sent TextEvent. */
     private final TreeSet<Integer> mPressedTextKeys;
@@ -178,6 +179,12 @@ public final class InputEventSender {
     public boolean sendKeyEvent(KeyEvent e) {
         int keyCode = e.getKeyCode();
         boolean pressed = e.getAction() == KeyEvent.ACTION_DOWN;
+
+        // Filter spurious key events from touchscreen hardware (e.g. Oplus/OnePlus
+        // touchpanel firmware injecting F-keys during touch/mouse operations).
+        if (filterTouchscreenKeyEvents && TouchInputHandler.sTouchpanelDeviceIds.contains(e.getDeviceId())) {
+            return true;
+        }
 
         if ((e.getFlags() & KeyEvent.FLAG_CANCELED) == KeyEvent.FLAG_CANCELED) {
             android.util.Log.d("KeyEvent", "We've got key event with FLAG_CANCELED, it will not be consumed. Details: " + e);
--- a/app/src/main/java/com/termux/x11/input/TouchInputHandler.java
+++ b/app/src/main/java/com/termux/x11/input/TouchInputHandler.java
@@ -42,7 +42,9 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.reflect.InvocationTargetException;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.Objects;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BiConsumer;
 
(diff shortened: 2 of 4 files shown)


DdogezD:
Hi @twaik , could you take a look at my pr when you have time? really appreciate your help, Thanks a lot for your work! :D

app/src/main/java/com/termux/x11/input/InputEventSender.java
```diff
@@ -179,6 +180,16 @@ public boolean sendKeyEvent(KeyEvent e) {
...
 
+        // Filter spurious key events from touchscreen hardware (e.g. Oplus/OnePlus
+        // touchpanel firmware injecting F-keys during touch/mouse operations)
+        if (filterTouchscreenKeyEvents && e.getDevice() != null) {
+            String deviceName = e.getDevice().getName();
+            if (deviceName != null && deviceName.toLowerCase().contains("touchpanel")) {
+                android.util.Log.d("KeyEvent", "Filtered touchscreen key event: keyCode=" + keyCode + " device=" + deviceName);
+                return true;
+            }
+        }
```
twaik: Doing so on every event seems to be CPU consuming a lot. Probably you should cache device number and renew it on every device rescan.

DdogezD:
Hi @twaik ,the issue you pointed out has been fixed. I verified the fix locally and force pushed a squashed commit.

twaik:
Reset existing hashset instead of creating new one.

DdogezD:
@twaik This should make more sense and match the overall structure of the code :D

twaik (approved):
LGTM.

twaik:
@robertkirkman do you have any thoughts?

app/src/main/java/com/termux/x11/input/TouchInputHandler.java
```diff
@@ -258,6 +263,9 @@ static public void refreshInputDevices() {
 
                     if (device.supportsSource(InputDevice.SOURCE_KEYBOARD) && device.getKeyboardType() == InputDevice.KEYBOARD_TYPE_ALPHABETIC && isExternal(device))
                         externalKeyboardAvailable.set(true);
+
+                    if (name != null && name.toLowerCase().contains("touchpanel"))
```
robertkirkman, line 267: I fully understand and sympathize with this problem because I work with an abnormal desktop-type device that has a touchpad on it, and its touchpad produces keyboard key events every single time a finger reaches the edge of the touchpad, no matter what, because the events are produced internally by the firmware of the touchpad, which is inaccessible.

on desktop linux running on that device, there is no way I know of to block the keyboard events from only that device globally.

however, if someone is using an abnormal device like that and having the same problem, but the device affected doesn't have "touchpanel" in the name, then this wouldn't help in that situation. I don't know yet whether the touchpad of my other device is named "touchpanel" but I am guessing it's not.

for that reason, if you make this configurable to show a menu of what devices are available and allow selecting enabling or disabling keyboard or touch or mouse event input for all of them, I would boot Android-x86 on that device and I would check whether this actually gives Termux:X11 the ability to function as a display server to solve this problem on that device, even when I don't know how to in GNU/Linux on the same device. and if it does then that would be great and I would probably switch to using Termux:X11 on that device to work around this issue when possible.

robertkirkman (commented):
I don't have a OnePlus device but I have a similar problem, could you make changes to help my device also?

app/src/main/java/com/termux/x11/input/TouchInputHandler.java
twaik: Can you please obtain termux-x11's debug log? It has device discovery logs.

DdogezD:
close this with https://github.com/termux/termux-x11/pull/1050, a better, more universal implementation.

PR #1050, opened 2026-07-06
feat(input): support per-device keyboard event blocking

### LINKED ISSUE #1051: [Bug]: Spurious key events injected by specific hardware
### Problem description

Spurious F4 key events injected by touchscreen hardware during touch/mouse operations.

The original build does log "Sending key" in logcat, which confirms that F4 key events (spurious key events) are being generated. However, it does not log which input device generated them.

By using PR https://github.com/termux/termux-x11/pull/1050, which adds a "recording" feature, we are able to determine whether there are any spurious key events and identify which device generated them. A screenshot is attached.

Image

Use a specific hardwares like my device “OnePlus Open(OxygenOS 16.0.5.701, android 16)” to use Termux x11.

Spurious key events get filtered.

DdogezD:
## Problem

Some device firmware (e.g. Oplus/OnePlus touchpanel) registers non-keyboard hardware as a keyboard input source, injecting spurious key events during touch and mouse operations. These reach the X server as real key presses, causing unexpected behavior in X11 applications.

## Solution

Add a "Block spurious key events" preference screen (Settings → Keyboard) that dynamically lists all keyboard-capable non-virtual input devices. Each device has a toggle to block its key events from reaching the X server. A "Record last key event" toggle at the top temporarily shows the most recent key each device sent as a summary, to help identify misbehaving devices.

## Implementation

- `TouchInputHandler.refreshInputDevices()` caches filtered device IDs, resolved from per-device boolean preferences on each device rescan
- `InputEventSender.sendKeyEvent()` checks the cached set before any key processing
- Recording is guarded by a cached boolean, updated on preference change and device rescan
- Device listing uses standard `SwitchPreferenceCompat` + `PreferenceDataStore` patterns

## Testing

Tested on ARM64 Android 15 device with chroot XFCE via Termux:X11. Spurious F4 key events are correctly filtered when the device is toggled on. Recording toggle correctly shows the last key sent by each device. Normal keyboard and IME input unaffected.

### DIFF
--- a/app/src/main/java/com/termux/x11/input/InputEventSender.java
+++ b/app/src/main/java/com/termux/x11/input/InputEventSender.java
@@ -11,12 +11,14 @@ import static com.termux.x11.input.InputStub.*;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
 import android.graphics.PointF;
+import android.view.InputDevice;
 import android.view.KeyEvent;
 import android.view.MotionEvent;
 
 import com.termux.x11.MainActivity;
 
 import java.util.List;
+import java.util.Set;
 import java.util.TreeSet;
 
 /**
@@ -27,6 +29,12 @@ public final class InputEventSender {
     private static final int XI_TouchBegin = 18;
     private static final int XI_TouchUpdate = 19;
     private static final int XI_TouchEnd = 20;
+
+    private static final Set<Integer> TOUCH_NAV_KEYCODES = Set.of(
+        KEYCODE_BACK,
+        KEYCODE_HOME,
+        KEYCODE_APP_SWITCH
+    );
 
     private final InputStub mInjector;
     private final float[] mappedPoint = new float[2];
@@ -184,6 +192,9 @@ public final class InputEventSender {
             return true;
         }
 
+        if (shouldBlockKeyEvent(e))
+            return true;
+
         // Events received from software keyboards generate TextEvent in two
         // cases:
         //   1. This is an ACTION_MULTIPLE event.
@@ -264,5 +275,19 @@ public final class InputEventSender {
 
         // We try to send all other key codes to the host directly.
         return mInjector.sendKeyEvent(scancode, keyCode, pressed);
+    }
+
+    // Some touchpanel/touchpad hardware occasionally reports spurious key
+    // events (e.g. unexpected function keys) alongside its normal touch
+    // input, likely due to a firmware or driver quirk in the underlying
+    // input device. Filter these out to avoid having them misinterpreted
+    // as real keyboard input.
+    private boolean shouldBlockKeyEvent(KeyEvent event) {
+        InputDevice device = event.getDevice();
+        int sources = device != null ? device.getSources() : 0;
+        return (sources & InputDevice.SOURCE_KEYBOARD) != 0
+                && (sources & (InputDevice.SOURCE_TOUCHSCREEN | InputDevice.SOURCE_TOUCHPAD)) != 0
+                && (sources & InputDevice.SOURCE_STYLUS) == 0
+                && !TOUCH_NAV_KEYCODES.contains(event.getKeyCode());
     }
 }


DdogezD:
Hi @robertkirkman, this should be a better solution to the issue addressed in PR https://github.com/termux/termux-x11/pull/1049⁠

twaik:
Probably it would be better to detect touch panels and touch pads and simply ignore events generated by them...

robertkirkman:
> Probably it would be better to detect touch panels and touch pads and simply ignore events generated by them...

but what if, hypothetically, someone has a thing which looks like a touch pad with buttons on the side but they only want to use the buttons and disable the touch pad part? this is what I worry about with automatic detection and limited options in the menu.

I will test what you asked about but it will take a long time.

DdogezD:
> Probably it would be better to detect touch panels and touch pads and simply ignore events generated by them...

My device’s `touchpanel_pen1` also seems to use the same `oplus_keyevent_notifier`, so a heuristic based on SOURCE_TOUCHSCREEN/SOURCE_TOUCHPAD + SOURCE_KEYBOARD could also misclassify it...

It’s getting quite complicated, might be better to keep it manual.

EDIT: I don’t have the paired stylus, so I’m unable to tell whether this is actually an issue or not.

twaik:
The solution must work fine out of the box, so heuristics is the only answer.
Please, post debug log containing reports for device discovery and faulty keys presses and we will see what's wrong together and how to detect it.

@robertkirkman can you please post debug log too?

robertkirkman:
my device affected by this problem has GNU/Linux installed so it's necessary to boot android on it and I do not have access to it all the time only sometimes so I will check this but it will take some time

DdogezD:
> The solution must work fine out of the box, so heuristics is the only answer. Please, post debug log containing reports for device discovery and faulty keys presses and we will see what's wrong together and how to detect it.
> 
> @robertkirkman can you please post debug log too?

I’m going to sleep now, but I will upload the debug logs in an issue tomorrow :D

DdogezD:
@twaik The fix works for me :D Let’s wait for @robertkirkman to test it on his device as well.

robertkirkman:
i will try to but currently i am working on these issues also:

- https://github.com/termux/termux-packages/issues/30453
- https://github.com/termux/termux-packages/pull/30359
- `scripts/run-docker.sh ./build-package.sh -I -f openfoam`

app/src/main/java/com/termux/x11/input/InputEventSender.java
```diff
@@ -184,6 +192,14 @@ public boolean sendKeyEvent(KeyEvent e) {
             return true;
         }
 
+        // Some touchpanel/touchpad hardware occasionally reports spurious key
+        // events (e.g. unexpected function keys) alongside its normal touch
+        // input, likely due to a firmware or driver quirk in the underlying
+        // input device. Filter these out to avoid having them misinterpreted
+        // as real keyboard input.
```
twaik: Move that comment down to `shouldBlockKeyEvent` func definition please.

twaik (changes requested):

twaik (approved):
LGTM. Waiting for review from @robertkirkman

robertkirkman:
I have not tested yet, but something I can say first is that the title of the PR is currently about "per-device toggle", but this is not a per-device toggle, it is currently an autodetection. i understand that it needs to be an autodetection for the needs of the project, but the title of the PR should match the content of the code.

app/src/main/java/com/termux/x11/input/InputEventSender.java
```diff
@@ -28,6 +30,12 @@ public final class InputEventSender {
     private static final int XI_TouchUpdate = 19;
     private static final int XI_TouchEnd = 20;
 
+    private static final Set<Integer> TOUCH_NAV_KEYCODES = Set.of(
+        KEYCODE_BACK,
+        KEYCODE_HOME,
+        KEYCODE_APP_SWITCH
```
robertkirkman, line 36: If I remember what happens correctly and assuming the key code it will translate to in Android will be equivalent, I believe that one of the keycodes that my abnormal touchpad device produces is `KEYCODE_DPAD_LEFT`, which isn't listed here, but I haven't tested it yet to confirm that

twaik:
@robertkirkman can you please test if your touchpad triggers spurious events with this pr?

robertkirkman:
I am still trying to fix `scripts/run-docker.sh ./build-package.sh -I -f openfoam` but when I have an opportunity I will go to the device that has the affected touchpad and install Android on it and test this, and I also have to wait until a time when I have access to that device for more than about 24 hours so that I can remove the SSD and replace it with a drive that has Android and then also remove Android again and put it back how it was with no changes.

robertkirkman:
I think that i might have solved `scripts/run-docker.sh ./build-package.sh -I -f openfoam` now and after finishing that I will try to check this

- https://github.com/termux/termux-packages/pull/30612

twaik:
@robertkirkman I hate to ping people but can you please test this when you have some time?

robertkirkman:
Yes i will but currently (as I'm writing) someone else is using the affected touchpad device and they use it with GNU/Linux, not Android, so I have to wait to borrow it from them until they are not using it

twaik:
Rebase to the latest master please.

robertkirkman:
Termux APK  is loadable on the device which has a touchpad problem for me, but Termux:X11 APK failed to load on it. I am continuing to work on trying to make Termux:X11 load on it, but that should not block this now because probably the device is not yet compatible with Termux:X11.

twaik:
@DdogezD ?

DdogezD:
> @DdogezD ?

will check the diff and rebase tomorrow, little bit busy today :D

robertkirkman:
> Termux APK is loadable on the device which has a touchpad problem for me, but Termux:X11 APK failed to load on it. I am continuing to work on trying to make Termux:X11 load on it, but that should not block this now because probably the device is not yet compatible with Termux:X11.

I installed an updated ROM on a device that had the same error and now it's not having the error anymore, so I'm going to try to get the updated ROM to the same device that has the touchpad which I could try to test with this.

twaik:
@DdogezD ?

PR #1052, opened 2026-07-07
perf(renderer): drop cond var relay thread, wake renderer directly

twaik:


twaik:
@robertkirkman mind taking a look? Went through reconnect/restart scenarios (including the X server process dying independently and the activity dying via force-stop) and couldn't find a way to break it. Curious if you spot something I'm missing.

robertkirkman (approved):
I tested this and for me it's working and I don't notice any problem.

robertkirkman:
If it is meant to be faster, that's good, because it doesn't feel slow. I don't know if it's faster than before but it's probably not slower.

twaik:
It is meant to optimize the hot path of renderer (waiting for new event notification) so it can reduce waiting time by a few ns for each frame. Not meant to be too much significant. But it could improve fps by 1-3 frames on high frequency displays so it should be good.

robertkirkman:
Ah, I don't have a high frequency display currently with android installed, but I have a question, is is feasible for the "Adjust display height for extra keys bar" option to be enabled by default now, if I ask for that, or does it have to stay disabled by default?

twaik:
Done.

PR #1064, opened 2026-07-26
refactor(lorie): rewrite renderer in a runtime-free C++ subset

twaik:
First step towards multi-instance support (multiple X servers/windows at once): renderer state currently lives in file-scope C globals, which assume exactly one renderer per process. This moves it into a single `Renderer` object instead, so it can eventually become one-per-instance — no functional change here, just groundwork.

Written in a runtime-free subset of C++ (no exceptions, no RTTI, no STL, `-nostdlib++`), so it adds no `libc++` dependency to `Xlorie.so`.

### DIFF
--- a/lorie/src/main/cpp/lorie/buffer.h
+++ b/lorie/src/main/cpp/lorie/buffer.h
@@ -3,6 +3,10 @@
 #include <linux/ashmem.h>
 #include <android/hardware_buffer.h>
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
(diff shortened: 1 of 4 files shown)


twaik:
Ended up looking more like a style pass than a rewrite into a C++ subset, lol - but structurally it is one: renderer state moved off file-scope globals into a single object, methods use `this` instead of an explicit context param. Built with `-fno-exceptions -fno-rtti -nostdlib++`, so it adds zero libc++ dependency to Xlorie.so. Should read almost line-by-line against master.

PR #1075, opened 2026-07-30
feat(shell-loader): generate the companion packages from gradle

twaik:
The wrapper scripts are templates now, filled in from shell-loader's own applicationId configuration, so a fork changes the package name in one place. buildCompanionPackage replaces build_termux_package and writes both archives with commons-compress, so dpkg-deb, tar and xz are no longer needed on PATH.


twaik:
@robertkirkman can you please test this (both deb and pacman packages, including generator cases)? Should make life a bit easier for forks and for termux-generator.

robertkirkman (approved):

PR #1084, opened 2026-08-01
build: add a flavor with termux's sharedUserId, replacing termux-app with x11 builtin

twaik:


robertkirkman:
Should the name of the workflow step in the `.yml` file that builds the normal, non-shared-uid APK be changed from "app-universal-debug" to just "universal-debug" because the name of the termux-x11 APK no longer starts with "app"?

image

twaik:
I do not think it is necessary.

PR #1090, opened 2026-08-05
feat(input): sync Caps/Num/Scroll Lock state from Android to X server

twaik:
Forwards the lock-key bits from every real KeyEvent/MotionEvent's metaState to the X server, so toggling them while termux-x11 was backgrounded (and unseen by us) gets picked up as soon as focus returns. The server applies the state directly to XKB (locked_mods for Caps/Num, the indicator for Scroll Lock, which isn't tied to a real modifier in stock keymaps) instead of emulating key presses, so no spurious KeyPress/KeyRelease reaches clients.


robertkirkman:
I have tested this and this is working for me

PR #1091, opened 2026-08-05
feat(power): replace keepScreenOn toggle with an idle timeout

twaik:
Replaces the boolean with Never/1/5/10/20/60 minute/Off options; timed modes release the flag after real input goes idle, letting the system's own timeout take over. Existing boolean values migrate automatically.


robertkirkman:
> timed modes release the flag after real input goes idle, letting the system's own timeout take over.

what does that mean, does it mean that if using Termux:X11 and KeepOn at the same time, the timeout from Termux:X11 and the timeout from KeepOn will stack on top of each other?

robertkirkman:
It stacks on top of the timeout from KeepOn; I set KeepOn to 5 minutes and Termux:X11 to 10 minutes and the screen stayed on in 2 tests for 15 minutes.

robertkirkman (approved):

twaik:
Probably should call the option "screen timeout" and subtract system's own screen-off timeout.

robertkirkman:
Yes, I think so, if it's not too difficult to get the system screen time out and display it in the setting status next to the Termux:X11 added timeout. KeepOn does and sets it also, but it seems, I think, that in Termux:X11 you want to read the system screen time out but not override it, but extend it with another timeout when Termux:X11 is focused.

https://github.com/twentynine78/KeepOn

robertkirkman:
though, the phrase "subtract system's own screen-off timeout" makes it seem like you might have read me saying "stacks on top of" to mean "times out concurrently"; I guess I meant "times out consecutively" rather than what I said.

twaik:
It is meant to not stack one timer after another, new version should fix that.

robertkirkman:
Now I set the system timeout to 5 minutes and Termux:X11 timeout to 10 minutes, and the screen stayed on for 10 minutes, and then I set the system timeout to 10 minutes and the Termux:X11 timeout to 5 minutes, and the screen also stayed on for 10 minutes, so now is whichever one is longer the one used? Is this how it is supposed to be?

twaik:
It is not intentional, but since we can not override system timeout I added a warning about system timeout taking priority.

robertkirkman:
> we can not override system timeout

except that there is an app that I linked above that does... that is why I used "KeepOn" and "system timeout" interchangeably - KeepOn stores its timeout in the system timeout variable.

twaik:
image
It requires special permissions.
I do not think adding these to X server app is a good idea.
We already have false positive av checks for both termux and termux-x11, I do not want to create even more of these.

lorie/src/main/java/com/termux/x11/LoriePreferences.java
```diff
@@ -371,6 +376,35 @@ void updatePreferencesLayout() {
...
+            if (systemTimeoutMs <= Long.parseLong(mode) * 60_000L) {
+                p.setSummary(p.getEntry());
+                return;
+            }
+
+            String warningText = getString(R.string.lorie_pref_summary_screenIdleTimeoutConflict, (systemTimeoutMs / 60_000) + " min");
+            SpannableString warning = new SpannableString(warningText);
+            warning.setSpan(new ForegroundColorSpan(0xFFFFA000), 0, warningText.length(), Spannable.SPAN_EXCLUSIVE_EXCLUSIVE);
+            warning.setSpan(new StyleSpan(Typeface.BOLD), 0, warningText.length(), Spannable.SPAN_EXCLUSIVE_EXCLUSIVE);
+            p.setSummary(TextUtils.concat(p.getEntry(), "\n", warning));
```
robertkirkman, line 425: Is it possible for "Auto" to be called "System (%s)" instead of "
Auto" and instead of "%s" display the current system time out, or is that too hard?

robertkirkman: I am asking because my first impression of this version is that it is not clear to a person who does not read the code and understand everything happening what "Auto" means exactly, and I believe "System (%s)" would be a significantly clearer representation of what the option does, is that right?

robertkirkman (approved):

lorie/src/main/java/com/termux/x11/LoriePreferences.java
robertkirkman: yes that's exactly what I hoped it would be like

PR #1092, opened 2026-08-05
refactor(lorie): rewrite cmdentrypoint.c in a runtime-free C++ subset

twaik:


twaik:
@robertkirkman can you please check this for regressions?

robertkirkman:
Yes

lorie/src/main/cpp/lorie/cmdentrypoint.cpp
```diff
@@ -493,62 +479,66 @@ void lorieUnregisterBuffer(LorieBuffer* buffer) {
...
-    jmethodID adoptFd = (*env)->GetStaticMethodID(env, ParcelFileDescriptorClass, "adoptFd", "(I)Landroid/os/ParcelFileDescriptor;");
+    jclass ParcelFileDescriptorClass = env->FindClass("android/os/ParcelFileDescriptor");
+    jmethodID adoptFd = env->GetStaticMethodID(ParcelFileDescriptorClass, "adoptFd", "(I)Landroid/os/ParcelFileDescriptor;");
     socketpair(AF_UNIX, SOCK_STREAM, 0, client);
-    QueueWorkProc(addFd, NULL, (void*) (int64_t) client[1]);
+    QueueWorkProc(+[](__unused ClientPtr pClient, void *closure) -> Bool {
+        InputThreadRegisterDev((int) (int64_t) closure, handleLorieEvents, nullptr);
+        conn_fd = (int) (int64_t) closure;
+        lorieActivityConnected();
+        return TRUE;
```
robertkirkman, line 499: Why does `addFd()` need to become a unnamed labmda function now and other similar single use functions converted to lambda functions? Did you do that on purpose and like that way more?

twaik: Yes, intentional. These were only pulled out as named `static` functions because C had no lambdas — `QueueWorkProc` needs a function pointer. Now that the file is C++, short one-off callbacks like these can be defined inline via `+[](...){...}` right at their single call site instead of forcing readers to jump to a separate definition elsewhere in the file. No behavior change, just less fragmentation. (`handleTouchEvent` stayed a named static since it's longer and not a trivial one-liner.)

robertkirkman (approved):
It is working and I don't think there there is something wrong with it.

PR #1093, opened 2026-08-09
refactor(lorie): tie native LorieView resources to the Java view's own lifecycle

twaik:
Native GUI-side state (renderer, connection fd, JNI refs) is bundled into a single LorieViewResources object, heap-allocated in LorieView's constructor and freed in onDetachedFromWindow, instead of living forever in scattered process globals. The renderer's background thread, EGL context, and GL resources are torn down and joined before the object is freed.

Every native call now takes an explicit pointer to its own instance instead of implicitly reaching a shared global, so each LorieView addresses its own native state. Callers outside LorieView keep their existing method names and signatures; the pointer is threaded through internally.

A step towards supporting multiple LorieView/MainActivity instances side by side (multi-window).

### DIFF
--- a/lorie/src/main/cpp/lorie/activity.cpp
+++ b/lorie/src/main/cpp/lorie/activity.cpp
@@ -15,15 +15,16 @@
 #include <linux/in.h>
 #include <arpa/inet.h>
 #include <poll.h>
+#include <new>
 #include "lorie.h"
 
 #pragma clang diagnostic ignored "-Wunknown-pragmas"
 #pragma ide diagnostic ignored "cppcoreguidelines-narrowing-conversions"
 #pragma ide diagnostic ignored "ConstantFunctionResult"
 #define log(prio, ...) __android_log_print(ANDROID_LOG_ ## prio, "LorieNative", __VA_ARGS__)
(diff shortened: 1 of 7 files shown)


twaik:
@robertkirkman pinging since this a move towards the multi-window feature you requested.

robertkirkman:
I did not really request it, it was other people, but something that does concern me is, what I do is run multiple window managers at once, but those other people actually said they didn't want that, they want something that is analogous to, for example, VMWare "Unity Mode"

image

but with X11 in the guest and SurfaceFlinger in the host instead of Windows Xp in the guest and Windows Vista in the host.

Does this also move towards allowing that, or does this PR mainly make changes with only what I use, multiple separate window managers in multiple separate X11 server instances, in mind?

robertkirkman:
I know you said that you do not like how implementing that would make all app icons group together into a single app icon on the recent apps and also on the taskbar of SurfaceFlinger window managers that have a taskbar unlike the normal SystemUI. However, what I am concerned about is that I don't know if there is a great benefit to implementing something that is not going to be able to support what users asked for (assuming this PR does not move towards that)

twaik:
I may try to implement wm integration to android, but it will be VERY limited. And in that case I must handle multiple activity instances so this pr is a move towards that too.

robertkirkman:
Ok

PR #1096, opened 2026-08-14
feat(app): add xstartupCommand preference for CmdEntryPoint's X session

### LINKED ISSUE #664: feature reqest: add startup script
It's very often the case that users want to always read a certain script when X starts (e.g., .xinitrc).  Other systems provide this functionality, and it would be nice if termux-x11 could, so that we don't have to supply one every time we start it up.

twaik:
Lets a CLI-launched termux-x11 session pick up a command configured in the app's settings, at lower priority than -xstartup/-- and higher than $TERMUX_X11_XSTARTUP. Passed to the native start() call over JNI; not settable via termux-x11-preference, only from the app's own settings screen.

CmdEntryPoint now reads its own shared_prefs file directly when possible, falling back to a broadcast round-trip (reusing the existing ACTION_START intent's binder) when running under a different UID.

Fixes #664.

### DIFF
--- a/lorie/src/main/aidl/com/termux/x11/ICmdEntryInterface.aidl
+++ b/lorie/src/main/aidl/com/termux/x11/ICmdEntryInterface.aidl
@@ -1,7 +1,10 @@
 package com.termux.x11;
 
+import android.os.Bundle;
+
 // This interface is used by utility on termux side.
 interface ICmdEntryInterface {
     ParcelFileDescriptor getXConnection();
     ParcelFileDescriptor getLogcatOutput();
+    oneway void setPreferences(in Bundle prefs);
 }
--- a/lorie/src/main/cpp/lorie/InitOutput.c
+++ b/lorie/src/main/cpp/lorie/InitOutput.c
@@ -99,6 +99,12 @@ static lorieScreenInfo lorieScreen = {
 }, *pvfb = &lorieScreen;
 static char *xstartup = NULL;
 static char **xstartupArgv = NULL;
+static char *xstartupPreference = NULL; // Set over JNI, lower priority than -xstartup/-- on the command line.
+
+void lorieSetXstartupPreference(const char *cmd) {
+    free(xstartupPreference);
+    xstartupPreference = cmd && *cmd ? strdup(cmd) : NULL;
+}
 
 // Owned by the activity process, handed to us over the connection socket. Points at a placeholder until
 // the first connection so callers don't need a NULL check.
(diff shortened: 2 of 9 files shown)


robertkirkman:
leaving note in public that in my opinion:

- the true fix for #664 would be providing optionally a package `pkg install xorg-xinit` which provides `startx` command
- that builds from patches onto upstream `xinit` from Freedesktop.org
- that makes Termux:X11 work with `startx`
- and which also invokes `xinit` in as close as possible a way to how upstream does
- and patches it to work with Termux:X11 by conditionally disabling any parts of its code which conflict or don't work with Termux:X11
- and which defaults to the contents of `$PREFIX/etc/X11/xinit/xinitrc`
- but which uses `~/.xinitrc` instead if it exists https://gitlab.freedesktop.org/xorg/app/xinit/-/blob/3ee72eac4d78b0fc2f2fe7d62dc1a16e3134e358/startx.in#L199 
- none of this has to replace or disable preexisting startup methods for Termux:X11 such as `termux-x11 -xstartup` or `TERMUX_X11_XSTARTUP`
- and should even be integrated with other code however you think is appropriate if I also agree

I strongly recommend as reference examples for the UX of how **PC users of `startx` command expect it to behave**:

- Arch Linux vanilla https://wiki.archlinux.org/title/Installation_guide https://wiki.archlinux.org/title/Xinit
- FreeBSD vanilla https://docs.freebsd.org/en/books/handbook/bsdinstall/ https://docs.freebsd.org/en/books/handbook/x11/


I will give source code of termux-os if you need it but if you just look at Arch Linux and FreeBSD and imagine the prefix at `/data/data/com.termux/files/usr` instead of `/usr` or `/usr/local` and the libc being bionic instead of glibc or BSD then that is really all termux-os is.

To make `startx` work with Termux:X11, the only thing further to be done in comparison to that, is to switch the X11 server from DRM-Xorg to Termux:X11 and apply any changes necessary to do that, 

I don't want the `startx` command to be a termux-os-only command which is why I am talking about this so that it is clear that one of the reasons I did not release the code of termux-os' `xorg-xinit` package yet is because I prefer if it publicly works with Termux:X11 first because Termux:X11 is much more portable and usable on real Android than termux-os' X11 server will be.

robertkirkman:
Also so that you know here's some more information: There's a ton of code in upstream's `startx` that is marked with "`Darwin`", however, I've used XFCE on MacOS with the MacOS X11 server XQuartz and unfortunately for me whenever I've done that the `startx` command was **not** working, I had to do other random things and a XQuartz icon appeared on my taskbar, when it does work, XFCE opens automatically on MacOS just from me clicking this icon:

image

and it frustrated me that `startx` does not work on MacOS, so I avoid using MacOS partly because of that,

This information means that while the "`Darwin`" marked code in upstream might not be used/functional, I want `startx` to work in Termux on Android anyway, so that in theory, someone who has the same obsession with `startx` on PC as me, but who _does not also necessarily already like Termux_, would get a **good first impression by `pkg install xorg-xinit` and `startx` working as expected from the experience of that person**.

robertkirkman:
In my opinion, the fact that original issue reporter mentioned ".xinitrc" is proof that I am not just a crazy person rambling about an obsolete file that nobody except me uses on PC. making a branch of the PC `startx` codepath using patches to support Termux:X11 would increase consistency between desktop Linux and Termux.

PR #1114, opened 2026-08-28
xserver: prevent repeated Present wait-fence callbacks

KiralyCraft:
This fixes a Termux:X11 crash by disarming a Present wait fence callback before reexecution so the same Present request cannot be invoked repeatedly. Required by https://github.com/lfdevs/mesa-for-android-container/pull/96.

### DIFF
--- a/lorie/src/main/cpp/patches/xserver.patch
+++ b/lorie/src/main/cpp/patches/xserver.patch
@@ -309,7 +309,16 @@
      present_check_flip2_ptr             check_flip2;
  
 +++ ./present/present_execute.c
-@@ -67,6 +67,14 @@
+@@ -34,5 +34,7 @@
+     ScreenPtr               screen = vblank->screen;
+     present_screen_priv_ptr screen_priv = present_screen_priv(screen);
+-
++    /* lorie: disarm, or miSyncTriggerFence()'s restart-scan re-fires this callback. */
++    present_fence_set_callback(vblank->wait_fence, NULL, NULL);
++
+     screen_priv->re_execute(vblank);
+ }
+@@ -67,6 +69,14 @@
      WindowPtr                   window = vblank->window;
      ScreenPtr                   screen = window->drawable.pScreen;
      present_screen_priv_ptr screen_priv = present_screen_priv(screen);
@@ -324,7 +333,7 @@
  
      /* If present_flip failed, we may have to requeue for the next MSC */
      if (vblank->exec_msc == crtc_msc + 1 &&
-@@ -79,12 +87,59 @@
+@@ -79,12 +95,59 @@
          return;
      }
  


KiralyCraft:
It looks like this upstream commit may be related: https://gitlab.freedesktop.org/xorg/xserver/-/commit/f19ab94ba9c891d801231654267556dc7f32b5e0

robertkirkman:
Is the bug this fixes reproducible without using proot-distro or chroot?

KiralyCraft:
In theory yes, since it is triggered by X11 Present and Sync requests handled inside Termux:X11. I currently only reproduced it using the #96 Mesa build from a chroot. A native Termux XCB client should be able to reproduce it by submitting PresentPixmap with an unsignaled wait fence and then triggering that fence while the request is queued for a future MSC; However, I did not yet run these tests yet. We are though following the specs.

The Present protocol explicitly allows a client to pass an unsignaled X Sync fence as PresentPixmap’s wait-fence. It requires the X server to defer presentation until that fence is triggered. Scheduling the same presentation for a future MSC is also valid, so the server must wait for both conditions. Present protocol specification ([see here](https://cgit.freedesktop.org/xorg/proto/presentproto/tree/presentproto.txt)) says "“PresentPixmap will block until 'wait-fence' is triggered."

The native Mesa build shipped by Termux does use DRI3 and Present, but it does not normally use this particular mechanism. It supplies no wait-fence and instead uses an idle-fence, which X triggers after it has finished accessing the pixmap so Mesa knows the buffer can be reused. The PR i referenced adds the opposite synchronization direction by converting the KGSL rendering completion fence into an X Sync wait-fence. Normal applications using the native Termux Mesa package are unlikely to use this exact callback path, although it remains a valid upstream bug imo.

robertkirkman:
What does #96 refer to?

robertkirkman:
oh I see you meant this https://github.com/lfdevs/mesa-for-android-container/pull/96

robertkirkman:
>  native Termux XCB client should be able to reproduce it by submitting PresentPixmap with an unsignaled wait fence and then triggering that fence while the request is queued for a future MSC; However, I did not yet run these tests yet. 

when you get time to, could you make and send this program?

KiralyCraft:
I built this minimal sample. It managed to crash the latest `termux-x11` from Releases as of today, in native Termux;

Compile with:
> clang -O00 minsrc.c -o minsrc     -lxcb-present -lxcb-sync -lxcb

Run with:
> DISPLAY=:0 ./minsrc

```
#include <xcb/xcb.h>
#include <xcb/present.h>
#include <xcb/sync.h>
#include <stdint.h>

int main()
{
    int screen_number;
    xcb_connection_t *c = xcb_connect(NULL, &screen_number);
    xcb_screen_iterator_t it = xcb_setup_roots_iterator(xcb_get_setup(c));
    while (screen_number-- != 0) xcb_screen_next(&it);
    xcb_screen_t *s = it.data;
    uint32_t one = 1;

    xcb_window_t w = xcb_generate_id(c);
    xcb_create_window(c, s->root_depth, w, s->root, 0, 0, s->width_in_pixels, s->height_in_pixels, 0, XCB_WINDOW_CLASS_INPUT_OUTPUT, s->root_visual, XCB_CW_OVERRIDE_REDIRECT, &one);
    xcb_map_window(c, w);
    xcb_present_select_input(c, xcb_generate_id(c), w, XCB_PRESENT_EVENT_MASK_COMPLETE_NOTIFY);
    xcb_pixmap_t p = xcb_generate_id(c);
    xcb_create_pixmap(c, s->root_depth, p, w, s->width_in_pixels, s->height_in_pixels);
    xcb_sync_initialize_reply(c, xcb_sync_initialize(c, 3, 1), NULL);
    xcb_sync_fence_t f = xcb_generate_id(c);
    xcb_sync_create_fence(c, w, f, 0);

    xcb_present_notify_msc(c, w, 1, 0, 0, 0);
    xcb_flush(c);
    xcb_present_complete_notify_event_t *e = (xcb_present_complete_notify_event_t *)xcb_wait_for_event(c);
    uint64_t target = e->msc + 60;
    xcb_present_pixmap(c, w, p, 2, XCB_NONE, XCB_NONE, 0, 0, XCB_NONE, f, XCB_NONE, 0, target, 0, 0, 0, NULL);
    xcb_present_notify_msc(c, w, 3, target - 1, 0, 0);
    xcb_flush(c);
    xcb_wait_for_event(c);

    xcb_sync_trigger_fence(c, f);
    xcb_flush(c);
    return xcb_wait_for_event(c) == NULL;
}
```

KiralyCraft:
Interestingly, it seems that `xcb_present_query_capabilities`  says `0x00000000`, and according to Present specs, using fences like this provides no benefit, "but also no cost". It does not say clients must not use them, but also says the driver must operate correctly without that capability, so I believe we're still "legal": [link to source of specs](https://github.com/ebe-forks/presentproto/blob/bfdc7e052302c79c5803ad95a73c9b63b350c40c/presentproto.txt#L202)

